Issue #41, opened 2014-12-21
Perspectives lost after last frame is closed when running in daemon mode


che2:
By default, perspectives are not shared between frames. To import a perspective to another frame,  the 'persp-import' command can be used.  This works as expected as long as the frame from which perspectives are improted is still open, but once that frame is closed, the perspectives belonging to the frame are lost forever, and the only way to recover the buffers from those lost frames are to either:
 o use 'persp-add-buffer' to add each buffer manually
 o shut down persp-mode

This is not ideal, especially in daemon mode, where the last frame would frequently be closed logging out of an ssh session, but perspectives are gone when a new frame is created in the same emacs daemon after logging back in. 

I suggest that either: 
- Add an option to share perspectives between all frames (which is done by https://github.com/Bad-ptr/persp-mode.el).
- Automatically transfer any perspectives to a closed frame to one of the remaining frames, and if the last frame is closed in daemon mode, the perspectives are automatically transfer to the first new frame is that is created from the same daemon.

Option 1 is probably the simplist to implement (it's already done by persp-mode), and also a good work around to the problem. Option 2 is arguably more sane, as it fixes the (broken) behavior.

nex3:
I like option one. In general, I think it's confusing if perspective data persists invisibly beyond the frame that owns them, but users who use daemonized instances a lot should have a way to persist perspectives.

I'd be open to merging in @Bad-ptr's implementation.

IvanMalison:
status? @che2 Would you be willing to incorperate the solution from @Bad-ptr ?

che2:
@IvanMalison, I am currently using #42 which implements solution 2 - transfering the workspaces of a frame when it is closed.  I am personally happy with this solution.

@nex3, would you be open to merge the open pull request?  I think it doesn't have drawbacks compared to the existing behavior:
- if you are using one frame for all your workspaces, then nothings changes
- if you are using multiple frames,  the current behavior is that closing one of the frames results in the loss of all workspaces belonging to that frame, which is very likely undesirable.  The new behavior is that the workspaces of that frame gets transferred to another live frame, which IMO is more desirable.  If a user doesn't like this behavior, he can set up a hook to kill all of the workspaces of a frame on close.

tmccombs:
I would like to have an option to share perspectives among all frames, as described in option one.

localredhead:
:+1:

nex3:
@che2 I've done a review of your PR.

Bad-ptr:
[...]

PR #42, opened 2014-12-21
Preserve perspectives when deleting a frame

che2:
Currently, when a frame is deleted, the perspectives associated with
that frame is lost.  This commit preserves the perspectives by
transferring them to another visible frame when available. Or saving the
perspectives from the last visible frame when emacs is running in daemon
mode so that they may be later restored.

Known Issues:
- When restoring the last visible frame in daemon mode, the previous
  activated perspective is not reactivated.  This is a deliberate
  choice, because saved window configurations does not transfer across
  frames ('persp-import' has the same problem), so reactivating the
  previously activated frame may inadvertently show a buffer from
  another perspective, which is undesirable.
#41

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -711,6 +711,7 @@ named collections of buffers and window configurations."
         (ad-activate 'exit-recursive-edit)
         (add-hook 'after-make-frame-functions 'persp-init-frame)
         (add-hook 'ido-make-buffer-list-hook 'persp-set-ido-buffers)
+        (add-to-list 'delete-frame-functions 'persp-frame-deleted)
         (setq read-buffer-function 'persp-read-buffer)
         (mapc 'persp-init-frame (frame-list))
         (setf (persp-buffers persp-curr) (buffer-list))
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -719,10 +720,43 @@ named collections of buffers and window configurations."
...
+      (persp-import-all-from-frame (get-other-frame) frame))))
+
+(defun persp-import-all-from-frame (to-frame from-frame)
+  "Import all perspectives from FROM-FRAME to TO-FRAME."
+  (let ((from-hash (with-selected-frame from-frame perspectives-hash))
+        (to-hash (with-selected-frame to-frame perspectives-hash)))
+    (cl-loop for name being the hash-keys of from-hash
+             when (not (gethash name to-hash)) do
+             (puthash name (gethash name from-hash)
+                      to-hash)))
```
nex3: Rather than just ditching the old perspective if there already exists one with the same name, I think we should merge them as best we can.

perspective.el
```diff
@@ -742,12 +783,13 @@ By default, this uses the current frame."
               (set-default 'header-line-format (append val '(persp-modestring)))))
         (setq global-mode-string (or global-mode-string '("")))
         (unless (memq 'persp-modestring global-mode-string)
-          (setq global-mode-string (append global-mode-string '(persp-modestring)))))
-      (persp-update-modestring))
```
nex3: Why did you remove this?

nex3:
Reviewed.

PR #43, opened 2014-12-23
Add persp-pick-perspective-by-buffer to switch perspectives by picking a

### LINKED ISSUE #30: Feature request: swich to perspective according to buffer
Hello,

thanks for sharing this useful package.

I wonder whether it could be possible to switch to a perspective by switching to a buffer.  Currently, you must switch perspective first and then buffer.  Being able to switch to a buffer directly would not only save typing, but also mental overhead.  This assumes that each buffer belongs to one perspective only, which should be the case when users are accustomed to use Perspective Mode.

Thanks for your attention.

IvanMalison:
buffer. Closes #30.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -130,7 +130,7 @@ them in Emacs >= 23.2.  In older versions, this is identical to
 
 (defvar persp-interactive-completion-function
   (if ido-mode 'ido-completing-read 'completing-read)
-  "The function which is used by perspective.el to interactivly complete user input")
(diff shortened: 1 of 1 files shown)


IvanMalison:
@nex3 Any chance you can look at this anytime soon?

IvanMalison:
@nex3 any chance of merge sometime soon?

IvanMalison:
bump.

PR #44, opened 2015-03-03
Support perspective wrapping on `persp-next` and `persp-prev`

redshirtrob:
Add customization point `persp-switch-wrap` to enable wrapping

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -68,6 +68,12 @@ perspectives."
                (string :tag "Close")
                (string :tag "Divider")))
 
+(defcustom persp-switch-wrap t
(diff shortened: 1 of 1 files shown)


nex3:
I'd be fine with this defaulting to `t`.

redshirtrob:
Same here, but I did not want to be presumptuous. :)  

I will fold that change in and update the PR.  Do you prefer that I squash the changes down to one commit?

nex3:
Yeah, that would be nice.

redshirtrob:
The PR is updated.

redshirtrob:
@nex3 I made the suggested change and updated the PR.

PR #45, opened 2015-03-25
Save `point-marker' as part of perspectives.

redshirtrob:


nex3:
I don't think this does anything on top of the existing functionality. The current active buffer and its point are properties of the window configuration, so they're already saved.

redshirtrob:
It does.  If you are visiting the same buffer in multiple windows then you will lose point in all but the current active buffer.

Try it out.

nex3:
Oh, I see.

PR #47, opened 2015-05-08
Respect ido's ignored files.

### LINKED ISSUE #28: ido-switch-buffer does not honor ido-ignore-buffers
When persp-mode is enabled any ignored buffers become visible. Doing an 'ido-switch-buffer' shows buffers like *Minibuf-x* and *Echo Area x*.

Ideally it would be possible to customize (and toggle) ido-ignore-buffers with persp-mode enabled.

nex3:
Closes #28

[Review on Reviewable](https://reviewable.io/reviews/nex3/perspective-el/47)

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -788,12 +788,12 @@ it. In addition, if one exists already, runs BODY in it immediately."
         (indices (make-hash-table :test 'equal)))
     (cl-loop for elt in ido-temp-list
              for i upfrom 0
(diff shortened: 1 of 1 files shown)


mromyers:
Is it possible to merge this? I've used this version for close to a year with no problems. If there's some problem I'm not aware of, sorry for dredging this back up.

PR #49, opened 2015-07-26
Define and run hook before switching perspectives

antonio:
This patch adds a new hook that is run just before a new perspective is activated.

I'm using it to save and restore perspectives automatically with [revive](https://github.com/emacsmirror/revive/).

```elisp
(defun antonio/save-perspective-configuration ()
  "Save the current perspective windows configuration"
  (interactive)
  (if persp-curr
      (with-temp-file (format "~/.emacs.d/perspectives/%s" (persp-name persp-curr))
        (insert (prin1-to-string (current-window-configuration-printable))))))

(defun antonio/load-perspective-configuration ()
  "Load the current perspective windows configuration"
  (interactive)
  (let ((perspective-file (format "~/.emacs.d/perspectives/%s" (persp-name persp-curr))))
    (if (f-exists? perspective-file)
        (restore-window-configuration (read (f-read perspective-file)))))))

(add-hook 'persp-before-switch-hook 'antonio/save-perspective-configuration)
(add-hook 'persp-created-hook 'antonio/load-perspective-configuration))
```

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -148,6 +148,10 @@ them in Emacs >= 23.2.  In older versions, this is identical to
   (if ido-mode 'ido-completing-read 'completing-read)
   "The function which is used by perspective.el to interactivly complete user input")
 
+(defvar persp-before-switch-hook nil
+  "A hook that's run before `persp-switch'.
+Run with the previous perspective as `persp-curr'.")
+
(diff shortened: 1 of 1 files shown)


antonio:
Thanks!

PR #52, opened 2015-10-16
advise set-window-buffer

csrhodes:
primarily for slime's benefit, which uses set-window-buffer to
replace the window displaying *inferior-lisp* with a slime
repl buffer (when the slime-repl contrib is active)


nex3:
Thanks!

PR #57, opened 2015-12-07
add persp-switch-last command

schouleu:
This command is used to switch to the last used perspective. This is
convenient when dealing often between two perspectives.
To access this command quickly, the binding is two times the prefix key.

Signed-off-by: Sylvain Chouleur <sylvain.chouleur@gmail.com>


nex3:
Thanks for the pull request!

Issue #64, opened 2016-12-21
FYI: Emacs 26 will remove frame-local variables


npostavs:
I believe this will break perspective-el.

https://lists.gnu.org/archive/html/emacs-devel/2016-12/msg00843.html

Gastove:
Just hit this after updating to a new nightly of Emacs 26; definitely breaks perspective.el.

nex3:
It should be possible to work around this by creating a map from frames to a struct that contains everything that's currently a frame-local variable. I'm not sure when I'll have time to do that, though.

redshirtrob:
It's less than ideal, but if you're not actually using multiple frames, you can add this:
```elisp
(when (not (fboundp 'make-variable-frame-local))
  (defun make-variable-frame-local (variable) variable))
```

john2x:
I did a bit of work in removing frame-local variables from the codebase in my fork at https://github.com/john2x/perspective-el.
I work with multiple frames extensively (via [nameframe](https://github.com/john2x/nameframe) package), so I needed perspective to be future proof :)

Initial tests seems good so far on Arch Linux, Emacs 25.3. Had a bit of trouble around the global-mode-string before I found out about `:eval` form. I'll do a bit more testing before opening a pull request.

nex3:
> I'll do a bit more testing before opening a pull request.

That would be great, thanks!

arichiardi:
Trying the @john2x branch as well (thanks!) and reporting back.

arichiardi:
@nex3 Don't know if this is the right place but with your fork I get an error:

```
Debugger entered--Lisp error: (void-variable perspectives-hash)
  (gethash name perspectives-hash)
  (let* ((name (file-name-nondirectory (directory-file-name project-to-switch))) (persp (gethash name perspectives-hash))) (cond ((and persp (not (equal persp persp-curr))) (persp-switch name)) ((not persp) (let ((frame (selected-frame))) (persp-switch name) (projectile-switch-project-by-name project-to-switch) (if (not (equal frame (selected-frame))) (progn (let ((old-frame (selected-frame)) (old-buffer (current-buffer))) (unwind-protect (progn (select-frame frame 'norecord) (persp-kill name)) (if (frame-live-p old-frame) (progn (select-frame old-frame 'norecord))) (if (buffer-live-p old-buffer) (progn (set-buffer old-buffer)))))))))))
  projectile-persp-switch-project("~/git/project/")
[...]

PR #67, opened 2018-05-29
Remove frame-local variables usage

john2x:
Hi. Now that Emacs 26.1 has been released, I thought it would be a good time to open a PR for my changes :) I've been using Emacs 26 pre-releases for a few months now (plus a few months more on Emacs 25) with my changes on perspective-el, and I haven't run into any issues.
(#64)

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -578,9 +578,9 @@ perspective and no others are killed."
     (setf (persp-killed (persp--current-persp)) t))
   (remhash name (frame-parameter nil 'perspectives-hash))
   (persp-update-modestring)
-  (when (equal name (persp-name (persp--last-persp)))
+  (when (and (persp--last-persp) (equal name (persp-name (persp--last-persp))))
     (set-frame-parameter nil 'persp-last nil))
-  (when (equal name (persp-name (persp--current-persp)))
+  (when (or (not (persp--current-persp)) (equal name (persp-name (persp--current-persp))))
     ;; Don't let persp-last get set to the deleted persp.
     (persp-let-frame-parameters ((persp-last (persp--last-persp)))
       (persp-switch (persp-find-some)))))
--- a/perspective.el
+++ b/perspective.el
@@ -196,54 +196,54 @@ Run with the activated perspective active.")
   (substitute-key-definition 'perspective-map nil persp-mode-map)
   (define-key persp-mode-map newkey 'perspective-map))
 
-;; make-variable-frame-local is obsolete according to the docs,
-;; but I don't want to have to manually munge frame-parameters
-;; all the time so I'm using it anyway.
-(make-variable-frame-local
- (defvar perspectives-hash nil
-   "A hash containing all perspectives. The keys are the
-perspectives' names. The values are persp structs,
-with the fields NAME, WINDOW-CONFIGURATION, BUFFERS,
-BUFFER-HISTORY, KILLED, POINT-MARKER, and LOCAL-VARIABLES.
+;; Keeping the variable definitions here since the docstrings might
+;; still be useful for developers.
 
-NAME is the name of the perspective.
+;; (make-variable-frame-local
+;;  (defvar perspectives-hash nil
+;;    "A hash containing all perspectives. The keys are the
+;; perspectives' names. The values are persp structs,
+;; with the fields NAME, WINDOW-CONFIGURATION, BUFFERS,
+;; BUFFER-HISTORY, KILLED, POINT-MARKER, and LOCAL-VARIABLES.
 
-WINDOW-CONFIGURATION is the configuration given by
-`current-window-configuration' last time the perspective was
-saved (if this isn't the current perspective, this is when the
-perspective was last active).
+;; NAME is the name of the perspective.
 
-BUFFERS is a list of buffer objects that are associated with this
-perspective.
+;; WINDOW-CONFIGURATION is the configuration given by
+;; `current-window-configuration' last time the perspective was
+;; saved (if this isn't the current perspective, this is when the
+;; perspective was last active).
 
-BUFFER-HISTORY is the list of buffer history values for this
-perspective.
+;; BUFFERS is a list of buffer objects that are associated with this
+;; perspective.
 
-KILLED is non-nil if the perspective has been killed.
+;; BUFFER-HISTORY is the list of buffer history values for this
+;; perspective.
 
-POINT-MARKER is the point position in the active buffer.
-Otherwise, when multiple windows are visiting the same buffer,
-all but one of their points will be overwritten.
+;; KILLED is non-nil if the perspective has been killed.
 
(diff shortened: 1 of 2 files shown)


perspective.el
```diff
@@ -196,54 +166,54 @@ Run with the activated perspective active.")
...
-;; but I don't want to have to manually munge frame-parameters
-;; all the time so I'm using it anyway.
-(make-variable-frame-local
- (defvar perspectives-hash nil
-   "A hash containing all perspectives. The keys are the
-perspectives' names. The values are persp structs,
-with the fields NAME, WINDOW-CONFIGURATION, BUFFERS,
-BUFFER-HISTORY, KILLED, POINT-MARKER, and LOCAL-VARIABLES.
+;; Keeping the variable definitions here since the docstrings might
+;; still be useful for developers.
```
nex3: Just get rid of these; this is why Git history exists.

perspective.el
```diff
@@ -293,9 +263,9 @@ for the perspective."
       (dotimes (_ 2) (push (pop args) keywords)))
     (setq keywords (reverse keywords))
     `(let ((persp (make-persp-internal ,@keywords)))
-       (when persp-curr
-         (setf (persp-local-variables persp) (persp-local-variables persp-curr)))
-       (puthash (persp-name persp) persp perspectives-hash)
+       (when (frame-parameter nil 'persp-curr)
```
nex3: Wrap `(frame-parameter nil 'persp-curr)` in a function so you don't have to repeat it all over the place. Same with other frequently-accessed frame parameters.

perspective.el
```diff
@@ -408,23 +378,28 @@ EVENT is the click event triggering this function call."
   (interactive "e")
   (persp-switch (format "%s" (car (posn-string (event-start event))))))
 
+(defun persp-mode-line ()
+  "Use this in the mode-line to show up-to-date value of `persp-modestring'."
```
nex3: `persp-modestring` doesn't refer to anything anymore. I'd write this docstring as "Return the string displayed in the modeline representing the perspectives."

Similarly, update other docstrings that refer to `persp-modestring`.

perspective.el
```diff
@@ -408,23 +378,28 @@ EVENT is the click event triggering this function call."
...
           (close (list (nth 1 persp-modestring-dividers)))
           (sep (nth 2 persp-modestring-dividers)))
-     (setq persp-modestring
+      (set-frame-parameter nil 'persp-modestring
            (append open
                    (persp-intersperse (mapcar 'persp-format-name
                                               (persp-names)) sep)
                    close)))))
 
+
```
nex3: Extra newline

perspective.el
```diff
@@ -618,35 +593,36 @@ See also `persp-switch' and `persp-add-buffer'."
...
-  (when (equal name (persp-name persp-last))
-    (setq persp-last nil))
-  (when (equal name (persp-name persp-curr))
+  (when (equal name (persp-name (frame-parameter nil 'persp-last)))
+    (set-frame-parameter nil 'persp-last nil))
+  (when (equal name (persp-name (frame-parameter nil 'persp-curr)))
+    (persp-switch (persp-find-some))
     ;; Don't let persp-last get set to the deleted persp.
-    (persp-frame-local-let ((persp-last persp-last)) (persp-switch (persp-find-some)))))
+    (set-frame-parameter nil 'persp-last (frame-parameter nil 'nil))))
```
nex3: I'm worried that setting `persp-last` to `nil` will break existing behavior. Can you set it to `(persp-find-some)` in its frame instead, as before?

perspective.el
```diff
@@ -781,22 +757,22 @@ See also `persp-add-buffer'."
...
-        ad-do-it
-        ;; We want the buffer lists that were created in the recursive edit,
-        ;; but not the window configurations
-        (maphash (lambda (key new-persp)
-                   (let ((persp (gethash key old-hash)))
-                     (when persp
-                       (setf (persp-buffers persp) (persp-buffers new-persp)))))
-                 perspectives-hash)
-        (setq perspectives-hash old-hash)))))
+    (set-frame-parameter nil 'persp-recursive (frame-parameter nil 'persp-curr))
```
nex3: Replacing `let` with `set` doesn't preserve the existing behavior--the frame parameter won't be reset after the function is called, or if it produces an exception.

I suggest replacing `persp-frame-local-let` with a new `persp-let-frame-parameter` macro that sets frame parameters for the duration of the body, and gracefully handles errors like `persp-frame-local-let`.

perspective.el
```diff
@@ -817,15 +793,18 @@ named collections of buffers and window configurations."
...
     (remove-hook 'after-make-frame-functions 'persp-init-frame)
     (remove-hook 'ido-make-buffer-list-hook 'persp-set-ido-buffers)
     (setq read-buffer-function nil)
-    (setq perspectives-hash nil)
-    (setq global-mode-string (delq 'persp-modestring global-mode-string))))
+    (set-frame-parameter nil 'perspectives-hash nil)
+    (setq global-mode-string (delete '(:eval (persp-mode-line)) global-mode-string))
+    (set-default 'header-line-format (delete '(:eval (persp-mode-line)) header-line-format))
+    (unless (delete "" header-line-format)
+      (set-default 'header-line-format nil))))
```
nex3: I don't understand what this statement is doing. Can you add a comment explaining?

perspective.el
john2x: I've removed the commented out variables.

perspective.el
john2x: I've added wrapper functions for getting `persp-curr` and `persp-last`.

perspective.el
john2x: I've updated the docstrings refering to `persp-modestring`,

perspective.el
john2x: I did some improvements around this, but I am not sure if it's the correct approach.

perspective.el
john2x: Added a new macro `persp-let-frame-parameters`, inspired by `persp-frame-local-let`.

perspective.el
john2x: Added a comment to this line. It seems that setting `header-line-format` to an empty string still renders it in the buffer. Setting it to nil removes it.

john2x:
Thanks for the review! I've pushed some changes to address the inline comments.

perspective.el
```diff
@@ -161,60 +173,19 @@ Run with the activated perspective active.")
 (define-key perspective-map (kbd "<left>") 'persp-prev)
 (define-key perspective-map persp-mode-prefix-key 'persp-switch-last)
 
+(defun persp--current-persp ()
```
nex3: Please make this and `persp-last` public. We want users to have access to this information.

john2x: Changed `persp--current-persp` and `persp--last-persp` to `persp-current-persp` and `persp-last-persp`.

PR #70, opened 2018-06-13
When removing a buffer, bury it in all windows displaying it

dzop:
Buries the removed buffer from all windows displaying it.


nex3:
Thanks!

PR #72, opened 2018-06-14
Handle an invalid `persp-point-marker` when activating a perspective

dzop:
This avoids an error that appears when the marker points nowhere. This can happen when the buffer that the marker originally pointed to was killed before switching back to the perspective being activated.


nex3:
Thanks!

PR #75, opened 2018-10-31
Prevent propagation of window parameters when switching perspectives

dzop:
`current-window-configuration` and `set-current-window-configuration` do not
care about window parameters so ensure that they are not in any of the windows
when creating new perspectives.

This is important because certain commands like `delete-other-windows` and
`split-window` take into account window parameters from newer versions of Emacs that
may inhibit these commands. The most notable one being `window-side`. We may want this behavior on the current
perspective, but not in any windows on the new perspective.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -334,6 +334,15 @@ REQUIRE-MATCH can take the same values as in `completing-read'."
            (when ,old (persp-switch ,old)))
          (set-frame-parameter nil 'persp--last last-persp-cache)))))
 
+(defun persp-reset-windows ()
+  "Remove all windows, ensure the remaining one has no window parameters.
+This prevents the propagation of reserved window parameters like
+window-side creating perspectives."
+  (let ((ignore-window-parameters t))
+    (delete-other-windows)
+    (when (ignore-errors (split-window))
+      (delete-window))))
+
 (defun persp-new (name)
   "Return a new perspective with name NAME.
 The new perspective will start with only an `initial-major-mode'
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -334,14 +334,23 @@ REQUIRE-MATCH can take the same values as in `completing-read'."
            (when ,old (persp-switch ,old)))
          (set-frame-parameter nil 'persp--last last-persp-cache)))))
 
+(defun persp-reset-windows ()
+  "Remove all windows, ensure the remaining one has no window parameters.
+This prevents the propagation of reserved window parameters like
+window-side creating perspectives."
+  (let ((ignore-window-parameters t))
+    (delete-other-windows)
+    (when (ignore-errors (split-window))
```
nex3: I'm not clear on what this conditional is doing.

dzop: That is a mystery. Calling `split-window` after calling `delete-other-windows` shouldn't error unless the frame is already too small to split in the first place which doesn't seem likely. 

It was clear to me when I wrote it a few months ago and I do remember running into errors without the call to `ignore-errors`, but that may have been due to some interaction with other packages I'm using. It does look like it is not needed.

On the other hand it does make it more robust on the off chance that the window split couldn't be made. Maybe use `split-window-no-error` to make it more clear what is intended? I'm fine with removing it also since it may have been due to interaction with my own packages and not a problem that `perspective` should deal with.

nex3: I'm partly just wondering why you need to run `split-window` at all.

dzop: The `delete-other-windows` removes all windows and leaves us in some window which may or may not have window parameters like `window-side`. The `split-window` call creates a fresh window without any window parameters, but we are still in the window that may or may not have window parameters we don't want in a new perspective, the `delete-window` call then ensures that the only window left is the new window without any window parameters.

nex3: Can you add a comment saying basically exactly that?

nex3:
Thanks!

PR #76, opened 2018-11-01
Keep track of the last time a perspective was switched to

dzop:
This is useful for displaying a time ordered list of perspectives when calling
`persp-switch` interactively. It also enables us to get rid of that annoying
"No last perspective" error when `persp-last` is killed.

* Add the `last-switch-time` field to a perspective which keeps track of the
  last time a perspective was switched to by the user.

* Add the `persp-prompt-chronologically` custom variable to possibly prompt for
  a perspective name showing them in an order according to their last access
  time.

* `persp-names` now takes an optional argument, `time-ordered`, when non-nil
  `persp-names` will return the names of the perspectives in an order according
  to their last switch time.

* When `persp-last` is killed, set it to the second most recently switched to
  perspective.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -83,6 +83,14 @@ perspectives."
   :group 'perspective-mode
   :type 'boolean)
 
+(defcustom persp-prompt-chronologically nil
+  "Whether to display perspective names in chronological order when prompting.
+If non-nil, show perspective names ordered by their last access
+time with most recently accessed first. Otherwise show names in
+alphabetical order."
+  :group 'perspective-mode
+  :type 'boolean)
+
 ;; This is only available in Emacs >23,
 ;; so we redefine it here for compatibility.
 (unless (fboundp 'with-selected-frame)
@@ -119,6 +127,7 @@ After BODY is evaluated, frame parameters are reset to their original values."
                (:conc-name persp-)
                (:constructor make-persp-internal))
   name buffers killed local-variables
+  (last-switch-time (current-time))
   (buffer-history buffer-name-history)
   (window-configuration (current-window-configuration))
   (point-marker (point-marker)))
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -466,6 +483,14 @@ perspective's local variables are set."
         (setq persp (persp-new name)))
       (run-hooks 'persp-before-switch-hook)
       (persp-activate persp)
+      ;; Only update the last switch time if one of the switching commands was
+      ;; explicitly called by the user.
+      (when (memq this-command '(persp-switch
+                                 persp-switch-last
+                                 persp-next
+                                 persp-prev
+                                 persp-switch-quick))
```
nex3: I worry that this isn't friendly to users who are defining their own interactive perspective commands. It also doesn't include e.g. `persp-mode-line-click`. I'd just make this `setf` part of the behavior of `persp-switch`.

dzop: Yes I was thinking about doing that, but what about `with-perspective`? Every time `with-perspective` is called, the perspective that was temporarily switched to would have a newer `last-switch-time` than the current perspective if we make the behavior part of `persp-switch`.

Maybe add a variable to temporarily disable setting the last switch time so the time doesn't get updated in `with-perspective` and possibly other places? Something like `persp-switch-norecord`?

dzop:
I think so. Then you could have a visual cue for which perspectives need some attention and which ones you visited last. I only added the behavior for `persp-prompt` since I only use `persp-switch` or `persp-last` to switch perspectives and only have the current perspective in the mode-line.

If the mode-line is considered, then it would also make sense to switch based on time with `persp-prev` and `persp-next`. So then, the behavior should probably be moved into `persp-names` as opposed to `persp-prompt` and rename `persp-prompt-chronologically` to `persp-sort-chronologically`. Then whenever `persp-sort-chronologically` is non-nil, `persp-names` returns a time ordered list. What do you think?

PR #80, opened 2019-02-25
Add support for saving and restoring perspectives to disk

gcv:
This is an attempt to add support for durable perspectives. I often have many open, with many buffers and windows in each. Restarting Emacs with so much state (for upgrades or just instability reasons) is really painful.

This has no dependencies aside from perspective-el itself. It provides no integration with desktop-save-mode, though I imagine few persp-mode users rely on it since it doesn't really do the right thing.

It works with my Emacs setup, but definitely needs testing with other people's configurations.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1128,6 +1128,9 @@ visible in a perspective as windows, they will be saved as
                (not (or current-prefix-arg
                         (yes-or-no-p "Target file exists. Overwrite? "))))
       (error "persp-state-save cancelled"))
+    ;; if ivy-posframe is in use, get rid of its frame first
+    (when (and (fboundp 'posframe-delete) (boundp 'ivy-posframe-buffer))
+      (posframe-delete ivy-posframe-buffer))
     ;; actually save
     (persp-save)
     (lexical-let ((state-complete (make-persp--state-complete
--- a/perspective.el
+++ b/perspective.el
@@ -1075,7 +1075,7 @@ transient."
                     persps-in-frame))))
 
 ;;;###autoload
-(defun persp-state-save (&optional file interactive?)
+(cl-defun persp-state-save (&optional file interactive?)
   "Save the current perspective state to FILE.
 
 FILE defaults to the value of persp-state-default-file if it is
@@ -1095,7 +1095,8 @@ visible in a perspective as windows, they will be saved as
                                 persp-state-default-file)
                 t))
   (unless persp-mode
-    (error "persp-mode not enabled, nothing to save"))
+    (message "persp-mode not enabled, nothing to save")
+    (return-from persp-state-save))
   (lexical-let ((target-file (if (and file (not (string-equal "" file)))
                                  ;; file provided as argument, just use it
                                  (expand-file-name file)
--- a/perspective.el
+++ b/perspective.el
@@ -177,13 +177,13 @@ Run with the activated perspective active.")
 (defvar persp-state-before-save-hook nil
   "A hook run immediately before saving persp state to disk.")
 
-(defvar persp-state-saved-hook nil
+(defvar persp-state-after-save-hook nil
   "A hook run immediately after saving persp state to disk.")
 
 (defvar persp-state-before-load-hook nil
   "A hook run immediately before loading persp state from disk.")
 
-(defvar persp-state-loaded-hook nil
+(defvar persp-state-after-load-hook nil
   "A hook run immediately after loading persp state from disk.")
 
 (defvar persp-mode-map (make-sparse-keymap)
@@ -1165,7 +1165,7 @@ visible in a perspective as windows, they will be saved as
       ;; create or overwrite target-file:
       (with-temp-file target-file (prin1 state-complete (current-buffer))))
     ;; after hook
-    (run-hooks 'persp-state-saved-hook)))
+    (run-hooks 'persp-state-after-save-hook)))
 
 ;;;###autoload
 (defun persp-state-load (file)
@@ -1232,7 +1232,7 @@ restored."
     ;; cleanup
     (persp-kill tmp-persp-name))
   ;; after hook
-  (run-hooks 'persp-state-loaded-hook))
+  (run-hooks 'persp-state-after-load-hook))
 
 (defalias 'persp-state-restore 'persp-state-load)
 
--- a/perspective.el
+++ b/perspective.el
@@ -1048,7 +1048,29 @@ potentially problematic buffers from the data structure created
 by window-state-get and replaces them with references to the
 perspective-specific *scratch* buffer. Buffers are considered
 'problematic' when they have no underlying file, or are otherwise
-transient."
+transient.
+
+The need for a recursive walk, and the consequent complexity of
+this function, arises from the nature of the data structure
+returned by window-state-get. That data structure is essentially
+a tree represented as a Lisp list. It can contain several kinds
+of nodes, including properties, nested trees representing window
+splits, and windows (referred to internally as leaf nodes).
+
+For the purposes of preserving window state, we only care about
+nodes in this data structure which refer to buffers, i.e., lists
+with the symbol 'buffer in the first element. These 'buffer lists
+can be deeply buried inside the data structure, because it
+recursively describes the layout of all windows in the given
+frame. They are always nested in lists with the symbol 'leaf in
+the first element.
+
+And so, the walker descends the data structure and preserves
+everything it finds. When it notices a 'leaf, it iterates over
+its properties until it finds a 'buffer. If the 'buffer points to
+a buffer which can be reasonably saved, it leaves it alone.
+Otherwise, it replaces that buffer's node with one which points
+to the perspective's *scratch* buffer."
   (cond
     ;; base case 1
     ((not (consp entry))
--- a/perspective.el
+++ b/perspective.el
@@ -174,6 +174,18 @@ Run with the perspective to be destroyed as `persp-curr'.")
   "A hook that's run after a perspective has been activated.
 Run with the activated perspective active.")
 
+(defvar persp-state-before-save-hook nil
+  "A hook run immediately before saving persp state to disk.")
+
+(defvar persp-state-saved-hook nil
+  "A hook run immediately after saving persp state to disk.")
+
+(defvar persp-state-before-load-hook nil
+  "A hook run immediately before loading persp state from disk.")
+
+(defvar persp-state-loaded-hook nil
+  "A hook run immediately after loading persp state from disk.")
+
 (defvar persp-mode-map (make-sparse-keymap)
   "Keymap for perspective-mode.")
 
@@ -1143,6 +1155,8 @@ visible in a perspective as windows, they will be saved as
                (not (or current-prefix-arg
                         (yes-or-no-p "Target file exists. Overwrite? "))))
       (error "persp-state-save cancelled"))
+    ;; before hook
+    (run-hooks 'persp-state-before-save-hook)
     ;; actually save
     (persp-save)
     (lexical-let ((state-complete (make-persp--state-complete
@@ -1150,7 +1164,9 @@ visible in a perspective as windows, they will be saved as
                                    :frames (persp--state-frame-data))))
       (when (file-exists-p target-file)
         (delete-file target-file))
-      (append-to-file (prin1-to-string state-complete) nil target-file))))
+      (append-to-file (prin1-to-string state-complete) nil target-file))
+    ;; after hook
+    (run-hooks 'persp-state-saved-hook)))
 
 ;;;###autoload
 (defun persp-state-load (file)
@@ -1169,6 +1185,9 @@ restored."
   (unless (file-exists-p file)
     (error "File not found"))
   (persp-mode 1)
+  ;; before hook
+  (run-hooks 'persp-state-before-load-hook)
+  ;; actually load
   (lexical-let ((tmp-persp-name (format "%04x%04x" (random (expt 16 4)) (random (expt 16 4))))
                 (frame-count 0)
                 (state-complete (read-from-whole-string
@@ -1212,7 +1231,9 @@ restored."
                                             (frame-root-window (selected-frame))
                                             'safe)))))
     ;; cleanup
-    (persp-kill tmp-persp-name)))
+    (persp-kill tmp-persp-name))
+  ;; after hook
+  (run-hooks 'persp-state-loaded-hook))
 
 (defalias 'persp-state-restore 'persp-state-load)
 
--- a/perspective.el
+++ b/perspective.el
@@ -1066,6 +1066,7 @@ transient."
 
 (defun persp--state-frame-data ()
   (cl-loop for frame in (frame-list)
+           if (frame-parameter frame 'persp--hash) ; XXX: filter non-perspective-enabled frames
            collect (with-selected-frame frame
                      (lexical-let ((persps-in-frame (make-hash-table :test 'equal))
                                    (persp-names-in-order (persp-names)))
@@ -1142,9 +1143,6 @@ visible in a perspective as windows, they will be saved as
                (not (or current-prefix-arg
                         (yes-or-no-p "Target file exists. Overwrite? "))))
       (error "persp-state-save cancelled"))
-    ;; if ivy-posframe is in use, get rid of its frame first
-    (when (and (fboundp 'posframe-delete) (boundp 'ivy-posframe-buffer))
-      (posframe-delete ivy-posframe-buffer))
     ;; actually save
     (persp-save)
     (lexical-let ((state-complete (make-persp--state-complete
--- a/perspective.el
+++ b/perspective.el
@@ -1012,13 +1012,16 @@ perspective beginning with the given letter."
 
 (defun persp--state-interesting-buffer-p (buffer)
   (and (buffer-name buffer)
-       (buffer-file-name buffer)
-       (not (string-match "^[[:space:]]*\\*" (buffer-name buffer)))))
+       (not (string-match "^[[:space:]]*\\*" (buffer-name buffer)))
+       (or (buffer-file-name buffer)
+           (with-current-buffer buffer (equal major-mode 'dired-mode)))))
 
 (defun persp--state-file-data ()
   (cl-loop for buffer in (buffer-list)
         if (persp--state-interesting-buffer-p buffer)
-        collect (buffer-file-name buffer)))
+        collect (or (buffer-file-name buffer)
+                    (with-current-buffer buffer ; dired special case
+                      default-directory))))
 
 (defun persp--state-window-state-massage (entry persp valid-buffers)
   "This is a primitive code walker. It removes references to
--- a/perspective.el
+++ b/perspective.el
@@ -994,10 +994,13 @@ perspective beginning with the given letter."
 ;;   :files [...]
 ;;   :frames [
 ;;     {
-;;       "persp1" {
-;;         :buffers [...]
-;;         :windows [...]
+;;       :persps {
+;;         "persp1" {
+;;           :buffers [...]
+;;           :windows [...]
+;;         }
 ;;       }
+;;       :order [...]
 ;;     }
 ;;   ]
 ;; }
@@ -1005,6 +1008,10 @@ perspective beginning with the given letter."
 (defstruct persp--state-complete
   files
   frames)
+
+(defstruct persp--state-frame
+  persps
+  order)
 
 (defstruct persp--state-single
   buffers
@@ -1059,23 +1066,27 @@ transient."
 
 (defun persp--state-frame-data ()
   (cl-loop for frame in (frame-list)
-        collect (with-selected-frame frame
-                  (lexical-let ((persps-in-frame (make-hash-table)))
-                    (cl-loop for persp in (persp-names) do
-                          (with-perspective persp
-                            (lexical-let* ((buffers
-                                            (cl-loop for buffer in (persp-buffers (persp-curr))
-                                                  if (persp--state-interesting-buffer-p buffer)
-                                                  collect (buffer-name buffer)))
-                                           (windows
-                                            (cl-loop for entry in (window-state-get (frame-root-window) t)
-                                                  collect (persp--state-window-state-massage entry persp buffers))))
-                              (puthash persp
-                                       (make-persp--state-single
-                                        :buffers buffers
-                                        :windows windows)
-                                       persps-in-frame))))
-                    persps-in-frame))))
+           collect (with-selected-frame frame
+                     (lexical-let ((persps-in-frame (make-hash-table :test 'equal))
+                                   (persp-names-in-order (persp-names)))
+                       (cl-loop for persp in persp-names-in-order do
+                                (unless (persp-killed-p (gethash persp (perspectives-hash)))
+                                  (with-perspective persp
+                                    (lexical-let* ((buffers
+                                                    (cl-loop for buffer in (persp-buffers (persp-curr))
+                                                             if (persp--state-interesting-buffer-p buffer)
+                                                             collect (buffer-name buffer)))
+                                                   (windows
+                                                    (cl-loop for entry in (window-state-get (frame-root-window) t)
+                                                             collect (persp--state-window-state-massage entry persp buffers))))
+                                      (puthash persp
+                                               (make-persp--state-single
+                                                :buffers buffers
+                                                :windows windows)
+                                               persps-in-frame)))))
+                       (make-persp--state-frame
+                        :persps persps-in-frame
+                        :order persp-names-in-order)))))
 
 ;;;###autoload
 (cl-defun persp-state-save (&optional file interactive?)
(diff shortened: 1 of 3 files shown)


nickgarber:
Just a note to say I'm interested in this feature and am testing it out as of a few moments ago. 
Happy to report on my experiences. :)

Really appreciate all the work that's gone into perspective, this feature included!

nickgarber:
I was getting errors from `persp-state-load` at `read-whole-buffer-string`.

It seemed like these records were accumulating and causing format issues on subsequent loads so I made two changes:
1) replace `read-from-whole-string` with `thing-at-point--read-from-whole-string` per the recommendation in function Help
2) delete the file prior to ever save


```
          (use-package perspective
            :after (projectile)
            :init
            (persp-mode)
            :config
            (progn
              ;;(straight-use-package '(perspective :type git :host github :repo "gcv/perspective-el"))
              (defun persp-persist-state-to-file () ""
                     (progn
                       (delete-file "~/.vault/cfgs/emacs/apps/perspective/persist.el")
                       (persp-state-save "~/.vault/cfgs/emacs/apps/perspective/persist.el")))
              (add-hook 'kill-emacs-hook 'persp-persist-state-to-file)
              (defun persp-resume-state-from-file ()  "" (persp-state-load "~/.vault/cfgs/emacs/apps/perspective/persist.el"))
              (add-hook 'after-init 'persp-resume-state-from-file)))

```

With those adjustments this seems to be working for me!

nickgarber:
@gcv, I'm curious to know if desktop-save interoperability/support on your list for future changes?

gcv:
@nickgarber: Thank you for testing!

I think I'll add support for a custom `persp-state-default-file` variable which, if set, will provide a default for `persp-state-save` and `persp-state-load`. Still need to think about what to do about deleting the target file if it exists before saving — leaning towards prompting the user first unless a prefix argument is used _or_ `persp-state-default-file` is set. That should make it possible to just add `persp-state-save` to `kill-emacs-hook` in your use case, but also avoid clobbering existing state files for people who might want to keep different persp state configurations.

I'm reluctant to use `thing-at-point--read-from-whole-string`, since (1) `read-from-whole-string` is an alias for it, and (2) it was only added in 25.1, and I'm not sure what versions of Emacs `perspective-el` officially supports.

About `desktop-save-mode` interoperability: I don't use it myself since it borks perspectives. I know it's supposed to help with font faces and variables, but I change those somewhat frequently so I don't find saving that helpful (different size screens, also I change through themes depending on ambient light). How do you use it, and how would you want it to integrate with `persp-state`?

nickgarber:
Your points in the first and second paragraph sound wholly sensible to me. 
I've adjusted my test case to use the below:

```
         (use-package perspective                                                                                                               
             :after (projectile)                                                                                                                  
             :init                                                                                                                                
             (persp-mode)                                                                                                                         
             :custom                                                                                                                              
             (persp-state-default-file "~/.vault/cfgs/emacs/apps/perspective/persist.el" "")                                                      
             :config                                                                                                                              
             (progn                                                                                                                               
               ;;(straight-use-package '(perspective :type git :host github :repo "gcv/perspective-el"))                                          
               (defun persp-persist-state-to-file () ""                                                                                           
                      (progn                                                                                                                      
                        (delete-file persp-state-default-file)                                                                                    
                        (persp-state-save persp-state-default-file)))                                                                             
               (add-hook 'kill-emacs-hook 'persp-persist-state-to-file)                                                                           
               (defun persp-resume-state-from-file ()  "" (persp-state-load persp-state-default-file))                                            
               (add-hook 'after-init 'persp-resume-state-from-file)))                                                                             
``` 

and (just a mention), I've also started testing this config alongside `psession` as:

```
           ;;; https://github.com/thierryvolpiatto/psession                                                                                       
           (use-package psession                                                                                                                  
             ;:disabled                                                                                                                           
             :init                                                                                                                                
             (require 'psession)                                                                                                                  
             (psession-mode 1)                                                                                                                    
             (psession-savehist-mode 1)                                                                                                           
             (psession-autosave-mode 1)                                                                                                           
             :hook                                                                                                                                
             ((emacs-kill . psession-save-winconf)                                                                                                
              (after-init . psession-restore-last-winconf))) 
```

and this all seems to be going well.


Perspective (+projectile) allows me to:
- maintain discrete and coherent projects
- lessen the cognitive load of switching between multiple projects over time

In particular, it's required to efficiently deliver on tasks that are prone to preemption.

When the `desktop-save` / session persistence functionality is added it just serves to extend the "over time" farther, across Emacs restarts.

Put simply, it means the mental safety-net can catch more preempted project work.

In the case of `desktop-save-mode` I believe I've also observed speed-ups when resuming vs when loading all of Emacs' init code, but thats anecdotal and secondary to the value described above.


Couple questions:

 - Q: Would it be possible to extend the work your doing to enable "hibernating" (and resuming) individual perspectives to disk? 

 - Q: Any suggestions or tips to make the interoperability between `psession` and `perspectives` more useful or profound?


Thanks!

gcv:
I just pushed some changes to support `persp-state-default-file`, and added documentation to the `README`. For your use case, just make sure that variable is set, and you can then point `kill-emacs-hook` to `persp-state-save` directly (i.e., you don't need your custom wrapper function, and can just use `(add-hook 'kill-emacs-hook #'persp-state-save)`).

Please test. 😁

It's definitely possible to extend `persp-state` to support saving and restoring individual perspectives. It would be a nifty feature, though not one I've personally needed so far. `persp--state-file-data` and `persp--state-frame-data` would need to learn to filter by perspective, and interactive entry point saving and loading functions would have to be added.

For psession integration: as far as I can tell, the only things it restores that `persp-state` does not are (1) buffer-local variables, and (2) registers. History is well-handled with `savehist-mode` in my setup. On the downside, psession will likely mess up at least the `main` perspective, because it'll open all files first, and then `persp-mode` will start and stuff them in `main`. `persp-state` is pretty careful to avoid doing that. Variables and registers don't matter in my workflow, so I don't really need psession. If I did need it, I think I'd hack it to make sure (1) it didn't open any files which are already open, and (2) make sure it runs _after_ `persp-state-load`, since by default it runs in `emacs-startup-hook`.

perspective.el
```diff
@@ -950,6 +960,209 @@ perspective beginning with the given letter."
...
+    (persp-save)
+    (lexical-let ((state-complete (make-persp--state-complete
+                                   :files (persp--state-file-data)
+                                   :frames (persp--state-frame-data))))
+      (when (file-exists-p target-file)
+        (delete-file target-file))
+      (append-to-file (prin1-to-string state-complete) nil target-file))))
+
+;;;###autoload
+(defun persp-state-load (file)
```
nickgarber: Should the `file` parameter be made `&optional`, to support using `persp-state-default-file` as the default/fall-back value?

gcv: It already defaults to `persp-state-default-file` in interactive mode. It can be made optional in non-interactive mode, but that would require pulling in all the extra dispatch machinery from `persp-state-save`. I guess it would be useful in `after-init-hook` to restore sessions after Emacs starts, but I personally wouldn't feel comfortable with doing that automatically (and it's not like someone might forget to restore the session, unlike at shutdown).

tbtommyb:
Hi from Reddit! I'm trying this out. I have set the `persp-state-default-file` to `~/.emacs.d/perspectives/store.el`. When I run `persp-state-save` I get:

```﻿
let*: Args out of range: #s(perspective "dotfiles" (#<buffer store.el> #<buffer *Messages*> #<buffer *Buffer List*> #<killed buffer> #<buffer init.el> #<buffer *Bookmark List*> #<buffer *Warnings*> #<buffer *GNU Emacs*> #<buffer perspective.el> #<killed buffer> ...) nil nil nil nil #<window-configuration>), 8
```
Can you advise? Thanks!

gcv:
Thank you for testing. I'll need your help to track this down.

Please evaluate the following code (using ielm for example):

```
(persp--state-frame-data)
```

Does that crash with the same args out of range? If it does then run this:

```
(cl-loop for buffer in (persp-buffers (persp-curr))
                                                  if (persp--state-interesting-buffer-p buffer)
                                                  collect (buffer-name buffer))
```

If all that succeeds, run this:

```
(make-persp--state-complete
                                   :files (persp--state-file-data)
                                   :frames (persp--state-frame-data))
```

PS: What version of Emacs?

tbtommyb:
Hello, apologies for the delay. I changed my config over to use `use-package` and wanted to get everything working before turning to this.

With my new config I can't recreate the error. I can try and use my old config if you want to track it down?

When I use `persp-state-save` and `persp-state-load` interactively things work well - thanks! There are a couple of points I noticed.

- my default Projectile action is to open dired in the project root. When I restore perspectives the dired buffers change to scratch. This is a bit annoying as scratch buffers seem to belong to the wrong project (i.e. `C-x C-f` from the scratch buffer takes me to a different project directory). This is a separate issue but it would be good if the restored perspectives included dired etc.
- When I run the commands interactively the default file I have specified appears at the top of the minibuffer, which looks a little odd. Is it meant to do that?

Screenshot 2019-08-04 at 20 54 30

- is there a way to control which perspective is active after loading? If I am in a perspective and load from file I would expect it to stay in the same perspective but with the restored arrangement. Or load the last active one on startup.

- finally I'm a bit unsure about how to run `persp-state-load` on startup because perspective etc only loads when I first run `C-c P P` due to use-package. Obviously that's nothing to do with the PR but what would you recommend?

Overall this works really well and is basically exactly what I was looking for so thanks for your work.

gcv:
Lots of good feedback here.

1. If you don't experience the problem any more, let's just let it go for now. I'm sure it'll come up again in some easier-to-reproduce situation.

2. Regarding dired: I wrote the code to explicitly ignore all non-file-visiting buffers because they often have non-reproducible state (e.g., buffers responsible for shells or other running processes). Obviously dired is a major exception here, and I had a feeling it would come up eventually :). Other modes come to mind, like treemacs, though that one will be harder to deal with. I should add special-case support at least for dired buffers.

3. I just call `read-file-name` with a default here. Judging by your screenshot, you use Ivy/Counsel — I can look into how they interact. Could you paste in your configuration for both Ivy and Counsel?

4. The random perspective order happens because I just write a (readable) hash table to disk. I can see how that can be annoying, and will look into how much work it would be to retain their order. That would be necessary for compatibility with #87 anyway.

5. Not sure, but I think adding `:demand t` to your `use-package` declaration will force `use-package` to eagerly load perspective. Or maybe you could just put a call to `(persp-state-load "...")` in your startup file and maybe that'll trigger the lazy load? Let me know if none of these work and I can try a little harder to think of a workaround. I use `use-package` myself but its lazy loading doesn't work quite right in my setup and I don't restart Emacs often enough to be annoyed and track it down.

I'm a bit swamped, but I'll try to find time to work on these improvements soon. Thanks for testing!

tbtommyb:
No worries! I will try those things for lazy loading. Here's my ivy/counsel config

```
(use-package counsel
  :ensure t
  )


(use-package ivy
  :ensure t
  :after counsel
  :config
  (ivy-mode t)
  (setq ivy-use-virtual-buffers t)
  ;; fuzzy matching
  (setq ivy-re-builders-alist '((t . ivy--regex-fuzzy)))
  (setq ivy-initial-inputs-alist nil)
  :bind (("C-s" . swiper)
         ("M-x" . counsel-M-x)
         ("C-c k" . counsel-ag)
         ("C-x C-f" . counsel-find-file)
         )
  )
```

gcv:
@tbtommyb: I just updated this PR with a commit enabling saving and restoring dired-mode buffers and windows. Seems to work for me. Could you please test it?

I looked into the Ivy strange display thing. It seems to be an artifact of the way `ivy-mode` modifies `read-file-name`. I guess I can detect ivy and handle it as a special case, but that workaround seems like a future maintenance hassle waiting to happen.

tbtommyb:
Yep, works nicely!

One other thing I've noticed is that I tend to leave lots of buffers open and also use flycheck. This means that restoring state can take quite a long time (15-20 seconds on a MBP). Maybe this is fine and inevitable because of how I do things. Personally I wouldn't expect perspective to reload anything more than the visible buffers but I guess it depends on how you use emacs. Anyway the functionality seems to be all there, thanks!

azzamsa:
@gcv sorry but why didn't we use desktop-mode functionality? ( I tried it but didn't work)

>  leave lots of buffers open and also use flycheck. This means that restoring state can take quite a long time (15-20 seconds on a MBP). @tbtommyb 

@gcv desktop-mode had `(setq desktop-restore-eager 5)`, so it will restore things lazily

``` txt
desktop-restore-eager’s value is t

Documentation:
Number of buffers to restore immediately.
Remaining buffers are restored lazily (when Emacs is idle).
If value is t, all buffers are restored immediately.
```

gcv:
desktop-mode predates perspective by a fair bit, and I don't see any non-fragile way to extend it. I'm certainly open to extending `persp-state-load` to load lazily, but I'm worried about introducing unpleasant corner cases.

azzamsa:
I haven't try you patch, but are you aware of shell buffers restoration?.
desktop-mode can't handle it, so we need desktop+ to restore shell buffers.

is this patch can restore shell buffers?

and, I saw some unmerged pr, are waiting for nex3 review/suggestion, or just waiting for tests?

gcv:
At this point, nex3 does not have time to work on this project. @zenspider and I have just received commit access, although we have not yet worked out the patch review process between us.

As for restoring shell buffers: the patch does not restore them at the moment, though I can probably add support for eshell specifically. Have to think about that one. Supporting term.el or vterm buffers is unlikely.

azzamsa:
> though I can probably add support for eshell specifically. Have to think about that one. Supporting term.el or vterm buffers is unlikely.

Is [shell](https://www.gnu.org/software/emacs/manual/html_node/emacs/Shell-Mode.html) will be supported, I use [shell-here](https://github.com/ieure/shell-here/blob/master/shell-here.el) all the time, really need this.

zenspider:
@azzamsa maybe I don't understand... It looks to me like `shell-here` opens a shell up relative to the buffer you're in... This is not exactly what `M-x shell` already does? I have some extra stuff to rename all my shell buffers to be unique so I can have many open, but otherwise ANY process is relative to the buffer you are in (or, that buffer's current value of `default-directory`)

gcv:
@azzamsa: I'm reluctant to add support for shell-mode, because it will be unreliable. The whole point of saving shell buffer state is to restore the shell with the same current working directory, but shell-mode does not have good directory tracking: https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Tracking.html#Directory-Tracking

The `M-x dirs` workaround fails if the cursor in the shell buffer is not at a shell prompt.

As I said, supporting eshell should not be a problem, as it does a good job of keeping `default-directory` up-to-date. Would you be able to switch to eshell for your use cases? In my opinion, shell-mode is the worst shell/terminal system Emacs has available. You could also try vterm, but it will not be supported with `persp-state-save` because it also has no directory tracking.

azzamsa:
@gcv okay great. https://github.com/ffevotte/desktop-plus maybe this project can give you some glimpse.

gcv:
@azzamsa: Thanks for the reference link. I looked at it. `desktop-plus` indeed relies on `default-directory` being set correctly, which makes sense but is not good. After a little experimentation, I managed to get shell-mode to set `default-directory` to truly bizarre values like `/../`. 😱

azzamsa:
>  managed to get shell-mode to set default-directory to truly bizarre values like /../. :scream:

Haha I hope that experiment didn't take much of your productive time. Just leave shell-mode support if that so difficult to deal with.

perspective.el
```diff
@@ -950,6 +962,236 @@ perspective beginning with the given letter."
...
+  (cond
+    ;; base case 1
+    ((not (consp entry))
+     entry)
+    ;; base case 2
+    ((atom (cdr entry))
+     entry)
+    ;; leaf: modify this
+    ((eq 'leaf (car entry))
+     (lexical-let ((leaf-props (cdr entry)))
```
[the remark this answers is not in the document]
gcv: Fair point, but I'd rather just convert perspective to lexical binding: https://github.com/nex3/perspective-el/issues/99

perspective.el
```diff
@@ -950,6 +962,236 @@ perspective beginning with the given letter."
...
+               (not (string-equal (if (and persp-state-default-file
+                                           (not (string-equal "" persp-state-default-file)))
+                                      (expand-file-name persp-state-default-file)
+                                    "")
+                                  target-file))
+               (not (or current-prefix-arg
+                        (yes-or-no-p "Target file exists. Overwrite? "))))
+      (error "persp-state-save cancelled"))
+    ;; if ivy-posframe is in use, get rid of its frame first
+    (when (and (fboundp 'posframe-delete) (boundp 'ivy-posframe-buffer))
```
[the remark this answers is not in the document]
gcv: It's definitely ugly to have this code in perspective. But if it burned me, it'll burn someone else, and I don't want to make people manually add these fixes. The right thing is to eliminate the use of frame parameters, and that will happen eventually. Meanwhile, though, the hook is a good idea. I'll add that, and add the ivy-posframe cleanup there by default.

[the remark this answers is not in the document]
gcv: I found a better and simpler workaround, which does not require posframe-specific code: 01e8370fcd

The hooks are a good idea regardless, so I added support for them: 8072c9f5cb

perspective.el
```diff
@@ -950,6 +962,236 @@ perspective beginning with the given letter."
...
+      (posframe-delete ivy-posframe-buffer))
+    ;; actually save
+    (persp-save)
+    (lexical-let ((state-complete (make-persp--state-complete
+                                   :files (persp--state-file-data)
+                                   :frames (persp--state-frame-data))))
+      (when (file-exists-p target-file)
```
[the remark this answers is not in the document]
gcv: Good call.

perspective.el
```diff
@@ -950,6 +962,236 @@ perspective beginning with the given letter."
...
+    (persp-switch tmp-persp-name)
+    (cl-loop for file in (persp--state-complete-files state-complete) do
+             (when (file-exists-p file)
+               (find-file file)))
+    ;; iterate over the frames
+    (cl-loop for frame in (persp--state-complete-frames state-complete) do
+             (incf frame-count)
+             (when (> frame-count 1)
+               (make-frame-command))
+             ;; backwards compatibility with older persp-state files: the frame
```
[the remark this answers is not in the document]
gcv: I agonized over this. The backwards compatibility was quite useful to me personally, and I've been suggesting that people use this unreleased patch, and several have kindly tested it. I'd rather not leave them with useless state files after this goes out. On the other hand, you're right: this adds ugly complexity.

In the end, I think this should go out with the backwards compatibility mode. It will help early adopters by not breaking their existing state files, and they will quickly generate new working state files as they use perspective. I'll remove the backwards compatibility layer in the (not distant) future, and left myself a comment as a reminder.

gcv:
Heads up for anyone following this PR: I rebased the underlying branch on top of the current mainline master and force-pushed. Reason: (1) some unrelated commits accidentally made it to this feature branch and I wanted to remove them, and (2) mainline recently acquired the ability to write automated tests, and I wanted to write tests for this feature before merging it in.

perspective.el
```diff
@@ -164,6 +174,18 @@ Run with the perspective to be destroyed as `persp-curr'.")
...
+  "A hook run immediately before saving persp state to disk.")
+
+(defvar persp-state-saved-hook nil
+  "A hook run immediately after saving persp state to disk.")
+
+(defvar persp-state-before-load-hook nil
+  "A hook run immediately before loading persp state from disk.")
+
+(defvar persp-state-loaded-hook nil
+  "A hook run immediately after loading persp state from disk.")
```
zenspider: Maybe `after-save` and `after-load`? (I'm biased towards symmetry)

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+  (cond
+    ;; base case 1
+    ((not (consp entry))
+     entry)
+    ;; base case 2
+    ((atom (cdr entry))
+     entry)
+    ;; leaf: modify this
+    ((eq 'leaf (car entry))
+     (lexical-let ((leaf-props (cdr entry)))
```
zenspider: It would help future maintainers if you documented/drew the shapes of things handled by this.

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+                                   prop
+                                 (cons 'buffer
+                                       (cons (lexical-let ((scratch-persp (format "*scratch* (%s)" persp)))
+                                               (if (get-buffer scratch-persp)
+                                                   scratch-persp
+                                                 "*scratch*"))
+                                             (cddr prop))))))))))
+    ;; recurse
+    (t (cons (car entry) (cl-loop for e in (cdr entry)
+                               collect (persp--state-window-state-massage e persp valid-buffers))))))
```
zenspider: This might just be my bias towards scheme/racket... but this (esp the `leaf` part) is really hard to follow. I'd want `match` for this function to make sense of it.

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+                                                    (cl-loop for entry in (window-state-get (frame-root-window) t)
+                                                             collect (persp--state-window-state-massage entry persp buffers))))
+                                      (puthash persp
+                                               (make-persp--state-single
+                                                :buffers buffers
+                                                :windows windows)
+                                               persps-in-frame)))))
+                       (make-persp--state-frame
+                        :persps persps-in-frame
+                        :order persp-names-in-order)))))
```
zenspider: sames... maybe I'm just agitating for a `TODO: refactor / clean?` comment?

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+'*scratch* (persp)' buffers."
+  (interactive (list
+                (read-file-name "Save perspective state to file: "
+                                persp-state-default-file
+                                persp-state-default-file)
+                t))
+  (unless persp-mode
+    (message "persp-mode not enabled, nothing to save")
+    (return-from persp-state-save))
+  (lexical-let ((target-file (if (and file (not (string-equal "" file)))
```
zenspider: Should we just do a full lexical conversion?

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+                               (if interactive?
+                                   ;; return nil in interactive call mode, since
+                                   ;; read-file-name should have provided a reasonable
+                                   ;; default
+                                   nil
+                                 ;; in non-interactive call mode, we want to fall back to
+                                 ;; the default, but only if it is set
+                                 (if (and persp-state-default-file
+                                          (not (string-equal "" persp-state-default-file)))
+                                     (expand-file-name persp-state-default-file)
```
zenspider: I think both these `if`s can reduce to a single 

```
(unless (and (not interactive?)
             persp-state-default-file
             (not blah...blah))
  ...)
```

zenspider: hrm... that logic is probably a tad wrong... but `(if cond1 nil (if cond2 result nil))` can definitely be reduced to `(when-or-unless cond3 result)`

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+                            (frame-persp-order (if (hash-table-p frame)
+                                                   (hash-table-keys frame)
+                                                 (reverse (persp--state-frame-order frame)))))
+               ;; iterate over the perspectives in the frame in the appropriate order
+               (cl-loop for persp in frame-persp-order do
+                        (lexical-let ((state-single (gethash persp frame-persp-table)))
+                          (persp-switch persp)
+                          (cl-loop for buffer in (persp--state-single-buffers state-single) do
+                                   (persp-add-buffer buffer))
+                          ;; XXX: split-window-horizontally is necessary for
```
zenspider: Please use "TODO" or "FIX" or something semantic.

perspective.el
```diff
@@ -968,6 +990,249 @@ perspective beginning with the given letter."
...
+                                                 (persp--state-frame-persps frame)))
+                            (frame-persp-order (if (hash-table-p frame)
+                                                   (hash-table-keys frame)
+                                                 (reverse (persp--state-frame-order frame)))))
+               ;; iterate over the perspectives in the frame in the appropriate order
+               (cl-loop for persp in frame-persp-order do
+                        (lexical-let ((state-single (gethash persp frame-persp-table)))
+                          (persp-switch persp)
+                          (cl-loop for buffer in (persp--state-single-buffers state-single) do
+                                   (persp-add-buffer buffer))
```
zenspider: Man.. I'm still not used to `cl-loop`...
```
(mapc #'persp-add-buffer (persp--state-single-buffers state-single)
```

seems clearer to me... but that's just me.

test/test-perspective.el
```diff
@@ -259,4 +259,23 @@ persp-test-make-sample-environment."
...
+  (unwind-protect
+      (persp-test-with-persp
+        (persp-test-with-files nil (A1 A2 A3 B1 B2 B3 B4)
+          (persp-test-make-sample-environment)
+          (should (= 7 (length (persp-test-buffer-list-all)))) ; sanity check
+          (persp-state-save "state-1.el"))
+        ;; reset perspectives
+        (persp-mode -1)
+        (delete-other-windows)
+        (should (= 0 (length (persp-test-buffer-list-all)))) ; no open files
```
zenspider: `seq-empty-p` seems clearer

perspective.el
gcv: Yes, that's what #99 is about. I'd rather do that as a separate step later.

perspective.el
gcv: XXX _is_ semantic. I use it to mean "misleading code which pretty much has to be that way for external reasons". For reference, I use FIXME to mean "must be fixed before shipping, and ideally ASAP", and TODO to mean "not ideal, but can ship as is (and therefore probably will never get fixed)".

I note that the [Wikipedia article on comments](https://en.wikipedia.org/wiki/Comment_%28computer_programming%29#Tags) has a similar explanation, so it's not just me. 🙃

perspective.el
gcv: Agreed: https://github.com/gcv/perspective-el/commit/ff7f03709f4edf445b841aff75cfad46dfd8da08

perspective.el
gcv: I expanded the docstring for the walker: https://github.com/gcv/perspective-el/commit/4a0fe25f037b453f0bf63f65aec20ec18ebd4cec

perspective.el
gcv: What is `match`?

gcv: This is a recursive tree walker which has to replace some of its nodes. Which is essentially a Lisp code walker. If there is a good abstraction built into Elisp for doing this, it can be used to implement the necessary functionality. Do you know of any such thing?

perspective.el
gcv: There was documented here: https://github.com/gcv/perspective-el/blob/4a0fe25/perspective.el#L1000-L1027

I'm certainly open to suggestions, but I don't see any way to simplify the code. It just takes Emacs' representation of window layouts, does a bit of transformation on unpersistable buffers, puts the result in a struct, and writes the struct to disk.

perspective.el
gcv: Right. I went through several iterations of writing this logic, and had to keep it as explicit as possible to get it right. There are too many combinations for how the function could be called: interactively or not, with a file argument or not, and what is an omitted file argument in interactive mode with a default available anyway?

I'm not happy with the look of the Boolean algebra here, but I've also been called out for writing overly-clever conditions in the past. 🤷🏻‍♂️

perspective.el
gcv: CL `loop` is extensively used throughout perspective. I just stuck with the existing convention.

test/test-perspective.el
gcv: Yea. But AFAIK seq.el was only added in Emacs 25.1, and we're still running automated tests going back to 24.3.

zenspider:
Match is amazing : https://docs.racket-lang.org/reference/match.html#%28form._%28%28lib._racket%2Fmatch..rkt%29._match%29%29

perspective.el
zenspider: There isn't one that I know of. `match` is amazing, but is a scheme thing.

zenspider:
Shit. Looks like my merges conflict with this. Sorry. I should have gotten this merged first since it is bigger / harder.

Once you rebase, please merge. We need this complexity in master and behind us.

gcv:
No big deal. It was a very minor conflict.

PR #85, opened 2019-05-13
Prevent pulling random buffers into current persp on buffer kill.

gcv:
There's currently an annoying bug when killing buffers after switching perspectives: sometimes, a buffer from a different perspective is pulled into the window from which the buffer was killed. I see that an attempt to fix this was made in 2015 with commit 2ccfe107d. Perhaps Emacs behavior changed since then, or perhaps it never quite worked.

The function advised in the previous fix attempt, `switch-to-prev-buffer`, switches to a buffer previously shown in the window. I observed that this can easily pull an unassociated buffer into the current perspective.

This PR includes a two-part fix. First, it makes sure that the `switch-to-prev-buffer` advice does not use the wrongly-updated current buffer list for its own scratch selection logic. Second, it adds its own advice to `window-prev-buffers` and `window-next-buffers` which filters their results by checking it against the list of buffers of the current perspective.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -783,16 +783,43 @@ See also `persp-add-buffer'."
         (with-selected-frame frame
           (persp-add-buffer buf))))))
 
+(defun persp--window-buffer-list-filter ()
+  (persp-protect
+    (when ad-return-value
+      (cl-loop with all = (persp-buffers (persp-curr))
+               for ble in ad-return-value
+               if (and ble
+                       (listp ble)
+                       (car ble)
+                       (not (null (buffer-name (car ble))))
+                       (memq (car ble) all))
+               collect ble into res
+               finally (setq ad-return-value res)))))
+
+(defadvice window-prev-buffers (after persp-window-prev-buffers)
+  "Ensure that any potential buffers to switch to belong to the
+  current perspective."
+  (persp--window-buffer-list-filter))
+
+(defadvice window-next-buffers (after persp-window-next-buffers)
+  "Ensure that any potential buffers to switch to belong to the
+  current perspective."
+  (persp--window-buffer-list-filter))
+
 (defadvice switch-to-prev-buffer (around persp-ensure-buffer-in-persp)
   "Ensure that the selected buffer is in WINDOW's perspective."
   (let* ((window (window-normalize-window window t))
          (frame (window-frame window))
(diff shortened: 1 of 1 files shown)


zenspider:
Can you describe a step-by-step reproduction? Either I don't understand this issue or I don't have it.

gcv:
Try this (without the patch):

1. Start with one perspective A.
2. Split it into a 2-3 windows, and visit a different file in each window.
3. Open another perspective B.
4. Split it into 2 windows, and visit a different file into each window (different files from perspective A).
5. Switch back to A.
6. Kill the current buffer (but not the current window).

This used to reliably put the last buffer I looked at from B into the current window of A.

dzop:
I can't reproduce this with your recipe both manually in my own Emacs instance and in `emacs -Q` with just `perspective` loaded, see #96.

I do recall experiencing this though, just can't remember what I did to make it happen.

There is the function `persp-reset-windows` that is supposed to have every new `perspective` start off in a fresh window that doesn't exist in any saved window configuration. That's what the call to `persp-reset-windows` is supposed to do in `persp-new`. This means that the `window-prev-buffers` of any window in a perspective never contains any buffers in other perspectives, except possibly if you call `persp-remove-buffer`. Maybe this is also important to reproduce the issue?

gcv:
Turns out I made this patch before you added `persp-reset-windows`, just didn't open a PR. The bug as described does not occur after commit 1358ba2c.

Also, I didn't describe it with enough detail to properly reproduce it. After reverting 1358ba2c (and without the patch in this PR), the following code reliably reproduces the bug:

```elisp
(progn
  (persp-switch "A")
  (select-window (split-window-right))
  (find-file "A1")
  (find-file "A2")
  (persp-switch "B")
  (select-window (split-window-right))
  (find-file "B1")
  (persp-switch "A")
  (persp-switch "B")
  (kill-buffer))
;; buffer "A1" will now be pulled into perspective B!
```

I'll close this ticket, but @dzop: please add this as a test case into #96, as it's much better than the one I described earlier.

dzop:
Done. And I can confirm that it does reproduce the issue when `persp-reset-windows` is a no-op, e.g. by evaluating `(advice-add 'persp-reset-windows :override #'ignore)` before running the test.

PR #87, opened 2019-05-27
Generalizes sorting of perspectives and adds sort-by-creation-time.

zenspider:
This is valuable to me because I often use perspectives during yak
shaves and other multi-project tasks where I need to "push" and "pop"
things to work on. By sorting by persp-time-created, this simulates
that perfectly:

+ While working on a task, find a tangent I need to go off on.
+ Create a new perspective.
+ Work on tangent. Possibly going back and forth between multiple perspectives.
+ When done with tangent, kill the perspective, going back to next-most-recent.

This switches from `persp-sort-chronologically` (bool) to
`persp-sort` (symbol). There might be some smoother way to honor the
old value and auto-convert to the new variable, but I don't know how.

This also adds `created-time` to the perspective struct. Loading this
code in your current emacs breaks everything. I don't know if it is
possible to migrate existing structs or otherwise patch this code to
work with old structs. Restarting emacs seems fine but I'm open to
suggestion.


zenspider:
I'm not thrilled with the way `persp-find-some` works when you're killing off perspectives and probably need to extend it for this change, but it's not really related to this change so much as it needs some more generalizations added to it.

The gist: Let's say you're using this mode to do a yak shave and you have several perspectives pushed on the stack. You finish up some of them and start to kill the perspectives. The first time, it uses `persp-last` to switch to the last one you were on. That seems fine. But after that, `persp-last` is `nil` and it goes to `(car (persp-names))` instead. That doesn't really match the workflow for this mode and should be something a bit smarter (newest, or `persp-last` should maintain a stack, or _something_).

zenspider:
> In the future, we can just recommend that people use persp-state-save before upgrading perspective-el.

does that have some sort of migration strategy or is it just flexible enough that it just fills in the blanks for fields that aren't there?

zenspider:
Merging...

zenspider:
fuck it. 😛

Issue #90, opened 2019-09-04
Make persp-find-some smarter


zenspider:
I'm not thrilled with the way `persp-find-some` works when you're killing off perspectives.

The gist: Let's say you're using this mode (persps sorted chronologically) to do a yak shave and you have several perspectives pushed on the stack. You finish up some of them and start to kill the perspectives. The first time, it uses `persp-last` to switch to the last one you were on. That seems fine. But after that, `persp-last` is `nil` and it goes to `(car (persp-names))` instead. That doesn't really match the workflow for this mode and should be something a bit smarter (newest, or `persp-last` should maintain a stack, or _something_).

_Originally posted by @zenspider in https://github.com/nex3/perspective-el/pull/87#issuecomment-528053148_

gcv:
I wonder if this might be the excuse needed to stop using frame parameters to store state and switch to a global hash table indexed by frame (id). Frame parameters cause other problems, like the one #86 attempts to address. It's kind of a big change, though.

gcv:
Looking at this and thinking about it... doesn't `(persp-names)` respect the sorting order? I just tried it, and perspective killing proceeded in pop-the-stack order. IOW Perspective currently behaves just like description suggests it should, at least for the yak shave workflow. Maybe I'm missing something?

gcv:
Let's reopen if this is still something we need to look at.

PR #91, opened 2019-09-04
Cleanup usage of frame parameters

dzop:
The main change of this pull request is to hide all modification of frame parameters behind the `setf` macro. So that instead of having `(set-frame-parameter nil 'persp-foo ...)` we have `(setf (persp-foo) ...)`. It also adds optional `frame` arguments to a few functions to avoid the use of `with-selected-frame` and also removes `persp-let-frame-parameters` as it seemed unnecessary.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -112,18 +112,6 @@ See also `with-temp-buffer'."
            (if (buffer-live-p ,old-buffer)
                (set-buffer ,old-buffer)))))))
 
-(defmacro persp-let-frame-parameters (bindings &rest body)
-  "Like `let', but for frame parameters.
-Temporariliy set frame parameters according to BINDINGS then eval BODY.
-After BODY is evaluated, frame parameters are reset to their original values."
-  (declare (indent 1))
-  (let ((current-frame-parameters (mapcar (lambda (binding) (cons (car binding) (frame-parameter nil (car binding)))) bindings)))
-    `(unwind-protect
-         (progn ,@(mapcar (lambda (binding) `(set-frame-parameter nil (quote ,(car binding)) ,(cadr binding))) bindings)
-                ,@body)
-       ;; Revert the frame-parameters
-       (modify-frame-parameters nil (quote ,current-frame-parameters)))))
-
 (cl-defstruct (perspective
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -482,7 +508,8 @@ If NORECORD is non-nil, do not update the
   (if (null name) (setq name (persp-prompt (and (persp-last) (persp-name (persp-last))))))
   (if (and (persp-curr) (equal name (persp-name (persp-curr)))) name
     (let ((persp (gethash name (perspectives-hash))))
-      (set-frame-parameter nil 'persp--last (persp-curr))
+      (when (and (persp-curr) (not (persp-killed (persp-curr))))
```
dzop: `(not (persp-killed (persp-curr)))` makes the `persp-let-frame-parameters` in `persp-kill` unnecessary.

perspective.el
```diff
@@ -790,42 +810,45 @@ See also `persp-add-buffer'."
...
+            (setq name (concat "*scratch*  (" (persp-name persp) ")")))
+          (with-selected-window window
+            (switch-to-buffer name)
+            (funcall initial-major-mode)))))))
 
 (defadvice recursive-edit (around persp-preserve-for-recursive-edit)
   "Preserve the current perspective when entering a recursive edit."
   (persp-protect
     (persp-save)
-    (persp-let-frame-parameters ((persp--recursive (persp-curr)))
```
dzop: We don't need this `persp-let-fram-parameters` since its usage is simple enough that we can just expand out what its doing.

dzop:
This is something that I've had in my local copy of `perspective` for almost a year, just never made a pull request, so I don't think there should be any issues it introduces.

gcv:
This is really interesting, thanks! I'll need more time to think about this patch.

There's an argument for eliminating frame parameters for maintaining perspective state altogether. In any case, I'd like to merge a few other PRs before this one and deal with potential conflicts in that order.

dzop:
Yes, I saw your comment in #90 which is why I thought it was the right time to make a pull request since it would make that transition extremely simple.

PR #95, opened 2019-09-19
Set live buffers as the `persp-buffers` when reactivating perspective 

dzop:
This should avoid the problem mentioned in #82 and also speed up reactivation of a long lived perspective that has seen many buffers.

It seems like the intention was to set `persp-buffers` on reactivation all along since `persp-reactivate-buffers` returns a list of live buffers. Nothing was ever done with the return value though.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -382,10 +382,10 @@ Returns BUFFERS with all non-living buffers removed.
 
 See also `other-buffer'."
   (cl-loop for buf in (reverse buffers)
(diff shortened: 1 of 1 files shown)


gcv:
I have not forgotten this PR. I want to test it now in my working session for a couple of days, and if I don't see any problems, will merge it in.

gcv:
This is great. It cleans up the internal state of perspective considerably.

Issue #89, opened 2019-08-22
Seems perspective is not working with emacs 26.2


ghost:
After installed perspective from MELPA, I added `(persp-mode)` to .emacs.
I got following error when opening emacs...
```
Symbol's function definition is void: hash-table-values
```
Seems this also breaks emacs and I am unable to close emacs or switch buffer.
I got following error when I tried to switch buffer.
```
persp-complete-buffer: Wrong type argument: perspective, nil
```

viveketic:
I don't think the issue is specific to emacs 26.2. I got same issue with emacs 25.2.2. I was able to solve the issue by adding following to `perspective.el` file. I am not sure this is the right way to fix the issue.
```
(eval-when-compile (require 'subr-x))
```
### References

- https://emacsredux.com/blog/2014/02/02/a-peek-at-emacs-24-dot-4-hash-table-keys-and-hash-table-values/

gcv:
This should be fixed in commit 7d3c966ee.

Issue #78, opened 2019-01-16
does not work with emacs 26.1


phenix3443:
(use-package perspective
  :ensure t
  :disabled
  :config
  (persp-mode)
  )

(use-package persp-projectile
  :after (projectile perspective)
  :disabled
  :ensure t
  :config
  ;; (define-key projectile-mode-map (kbd "s-s") 'projectile-persp-switch-project)
  )

when start emacs, show error:
Error (frameset): Wrong type argument: perspective, "Unprintable entity"

dzop:
I'm on

```
GNU Emacs 26.1.92 (build 2, x86_64-apple-darwin18.2.0, Carbon Version 158 AppKit 1671.2) of 2019-04-13
```

and I don't experience such an error. Also `frameset` functions are never called by perspective which makes me suspect that the error your are getting has to do with some other package interacting badly with perspective.

gcv:
@phenix3443: Any idea what other package might be causing perspective incompatibility? As things stand, packages which create frames with `after-make-frame-functions` bound to `nil` will cause all kinds of interesting perspective bugs. We'll try to resolve that soon.

gcv:
Closing this ticket, since we now have a test suite and cannot reproduce it. Please reopen if you still have problems.

PR #96, opened 2019-09-19
Add test framework

dzop:
I wasn't able to reproduce the issue mentioned in #85 with my local Emacs instance so I tried to reproduce it with `emacs -Q`. From there I just needed to add in some boiler plate to setup a test framework with Travis, so I went ahead and did that.

This should make it easier to ensure that no regressions are introduced as `perspective` gets more complex as well as provably show that `perspective` works as intended (we will be able to get rid of issues like #78 and #89).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+sudo: required
+# for evm binaries
+dist: trusty
+env:
+  - EMACS_VERSION=26.2
+  - EMACS_VERSION=26.1
+  - EMACS_VERSION=25.1
+  - EMACS_VERSION=24.3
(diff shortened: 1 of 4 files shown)


dzop:
I've never tried to setup CI on someone else's repo, I think either @nex3 or one of the maintainers needs to go to the Travis website and enable it.

cc @gcv @zenspider

gcv:
Great stuff.

I think this needs owner-level permissions to enable. Let me get this figured out.

gcv:
Tests enabled: https://travis-ci.org/nex3/perspective-el

PR #97, opened 2019-09-25
insert initial scratch message into new perspectives

npjg:
I have my initial scratch message globally set `nil`, as is sensible, but I often have Org boilerplate in perspective scratch buffers, which I can then save as attachments to the corresponding tasks. Since `perspective-el` supports `initial-major-mode`, it makes sense to also support `initial-scratch-message` for purposes like these.


gcv:
Seems reasonable on the whole, but I don't understand your workflow. How does setting the initial scratch buffer content help you save content — which you presumably have to write first — as Org attachments?

PR #102, opened 2019-10-19
Version bump

zenspider:
This should start to be the regular process when we merge PRs and get to stable checkpoints.


zenspider:
This should go in AFTER #103 since it completes something that's already shipped.

gcv:
Excellent!! First new release in a while!

PR #103, opened 2019-10-20
Extend persp-find-some to respect current perspective ordering when persp-last is nil

zenspider:
This completes the functionality I need to have `'created` act like a real work stack.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -560,6 +560,7 @@ This function tries to return the \"most appropriate\"
 perspective to switch to.  It tries:
 
   * The perspective given by `persp-last'.
+  * The \"first\" perspective, based on the ordering of persp-names.
   * The main perspective.
   * The first existing perspective, alphabetically.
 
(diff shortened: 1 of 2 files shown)


perspective.el
```diff
@@ -567,6 +567,7 @@ If none of these perspectives can be found, this function will
 create a new main perspective and return \"main\"."
   (cond
    ((persp-last) (persp-name (persp-last)))
+   ((> (length (persp-names)) 1) (car (persp-names)))
```
gcv: Please update the docstring of `persp-find-some` before merging.

perspective.el
```diff
@@ -567,6 +567,7 @@ If none of these perspectives can be found, this function will
...
    ((gethash persp-initial-frame-name (perspectives-hash)) persp-initial-frame-name)
    ((> (hash-table-count (perspectives-hash)) 0) (car (persp-names)))
```
zenspider: Is this one redundant at this point? Can it ever hit?

PR #104, opened 2019-10-22
Fix uninitialized frame crashes

gcv:


zenspider:
Can you 'splain _how_ you get this to occur? I don't hit this ever.

gcv:
Start from a fresh Emacs instance. Make sure you don't have this patch or #86 loaded.

Evaluate:

```
(let ((after-make-frame-functions nil))
  (make-frame))
```

This pops open a new frame. Perspective is not enabled in this frame, and will break if you attempt to use it. This is expected behavior (although far from ideal), as `after-make-frame-functions` bound to `nil` suppresses the hooks which handle the creation of all frame parameters which perspective relies on.

Unfortunately, this _also_ breaks perspective in the original frame in which all frame parameters are properly loaded. Keep the new frame open, go back to the original frame, make a new perspective, and then attempt to kill that new perspective. You'll get `Wrong type argument: hash-table-p, nil`.

This matters because [posframe](https://github.com/tumashu/posframe) does exactly this when creating its frames. posframe-dependent packages create hidden frames, none of which have the necessary frame parameter initialization. Most notably, this breaks [ivy-posframe](https://github.com/tumashu/ivy-posframe). To add insult to injury, this is not debuggable with `toggle-debug-on-error`.

I wrote a test, but I can't get it to work properly in terminal mode Emacs: calling `make-frame` in a batch test run errors out with "Unknown terminal type".

```
(ert-deftest issue-104-errors-with-uninitialized-frames ()
  (persp-test-with-persp
   ;;(persp-switch "A")
   (let ((after-make-frame-functions nil))
     (let ((orig-frame (selected-frame))
           (new-frame (make-frame)))
       (with-selected-frame orig-frame
         (persp-switch "A")
         (persp-switch "main")
         ;; should not error out:
         (persp-kill "test"))))))
```

This PR is an alternative to #86, which — as you noted — is a much more invasive approach to solving this problem. (It also complements the overall cleanup of frame parameter use in #91, which I haven't had time to test.)

gcv:
@zenspider: Any thoughts on this?

zenspider:
no thoughts. just merged. :P

PR #107, opened 2019-11-08
Fix trivial package-lint issues

conao3:
Hi!
I fix package-lint/Elisp-doc warned issues.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -6,7 +6,7 @@
 
 ;; Author: Natalie Weizenbaum <nex342@gmail.com>
 ;; URL: http://github.com/nex3/perspective-el
-;; Package-Requires: ((cl-lib "0.5"))
+;; Package-Requires: ((emacs "24.4") (cl-lib "0.5"))
 ;; Version: 2.3
 ;; Created: 2008-03-05
 ;; By: Natalie Weizenbaum <nex342@gmail.com>
--- a/perspective.el
+++ b/perspective.el
@@ -71,7 +71,7 @@ perspectives."
                (string :tag "Divider")))
 
 (defcustom persp-mode-prefix-key (kbd "C-x x")
-  "Prefix key to activate perspective-map"
+  "Prefix key to activate perspective-map."
   :group 'perspective-mode
   :set (lambda (sym value)
          (when (and (bound-and-true-p persp-mode-map)
@@ -86,15 +86,17 @@ perspectives."
   :type 'boolean)
 
 (defcustom persp-sort 'name
-  "What order to sort perspectives. If 'name, then sort alphabetically. If 'access, then sort by last time accessed (latest first). If 'created, then sort by time created (latest first)."
+  "What order to sort perspectives.
+If 'name, then sort alphabetically.
+If 'access, then sort by last time accessed (latest first).
+If 'created, then sort by time created (latest first)."
   :group 'perspective-mode
   :type '(choice (const :tag "By Name"          name)
                  (const :tag "By Time Accessed" access)
                  (const :tag "By Time Created"  created)))
 
 (defcustom persp-state-default-file nil
-  "When non-nil, it provides a default argument for
-`persp-state-save` and `persp-state-load` to work with.
+  "When non-nil, it provides a default argument for `persp-state-save` and `persp-state-load` to work with.
 
 `persp-state-save` overwrites this file without prompting, which
 makes it easy to use in, e.g., `kill-emacs-hook` to automatically
@@ -149,7 +151,7 @@ After BODY is evaluated, frame parameters are reset to their original values."
 
 (defvar persp-interactive-completion-function
   (if ido-mode 'ido-completing-read 'completing-read)
-  "The function which is used by perspective.el to interactivly complete user input")
+  "The function which is used by perspective.el to interactivly complete user input.")
 
 (defvar persp-before-switch-hook nil
   "A hook that's run before `persp-switch'.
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -86,15 +86,18 @@ perspectives."
...
   :group 'perspective-mode
   :type '(choice (const :tag "By Name"          name)
                  (const :tag "By Time Accessed" access)
                  (const :tag "By Time Created"  created)))
 
 (defcustom persp-state-default-file nil
-  "When non-nil, it provides a default argument for
-`persp-state-save` and `persp-state-load` to work with.
+  "When non-nil, it provides a default argument.
+For `persp-state-save` and `persp-state-load` to work with.
```
zenspider: this doesn't make sense to me

perspective.el
```diff
@@ -208,10 +211,11 @@ Run with the activated perspective active.")
...
   "Return a hash containing all perspectives in FRAME.
-FRAME defaults to the currently selected frame. The keys are the
-perspectives' names. The values are persp structs, with the
-fields NAME, WINDOW-CONFIGURATION, BUFFERS, BUFFER-HISTORY,
-KILLED, POINT-MARKER, and LOCAL-VARIABLES.
+FRAME defaults to the currently selected frame.
+The keys are the perspectives' names.
+The values are persp structs, with the fields NAME,
+WINDOW-CONFIGURATION, BUFFERS, BUFFER-HISTORY, KILLED,
+POINT-MARKER, and LOCAL-VARIABLES.
```
zenspider: Are you just re-wrapping to a narrower column? If so, why? To avoid double-spaces after periods? (which, I disagree with).

zenspider: add to custom: `(sentence-end-double-space nil)`

perspective.el
```diff
@@ -271,9 +275,10 @@ This prevents the persp-mode from completely breaking Emacs."
         (persp-mode -1)))))
 
 (defun persp-error (&rest args)
-  "Like `error', but marks it as a persp-specific error.
+  "Like `error', but mark it as a persp-specific error.
 Used along with `persp-protect' to ensure that persp-mode doesn't
-bring down Emacs."
+bring down Emacs.
+ARGS will interpreted by `format-message'."
```
zenspider: also strange formatting

perspective.el
```diff
@@ -1176,7 +1184,7 @@ visible in a perspective as windows, they will be saved as
                                   target-file))
                (not (or current-prefix-arg
                         (yes-or-no-p "Target file exists. Overwrite? "))))
-      (error "persp-state-save cancelled"))
+      (error "Persp-state-save cancelled"))
```
zenspider: I disagree with this one since it is a literal function name.

perspective.el
```diff
@@ -1043,34 +1052,33 @@ perspective beginning with the given letter."
...
-everything it finds. When it notices a 'leaf, it iterates over
-its properties until it finds a 'buffer. If the 'buffer points to
-a buffer which can be reasonably saved, it leaves it alone.
-Otherwise, it replaces that buffer's node with one which points
-to the perspective's *scratch* buffer."
+everything it finds.  When it notices a 'leaf, it iterates over its
+properties until it finds a 'buffer.  If the 'buffer points to a buffer
+which can be reasonably saved, it leaves it alone.  Otherwise, it
+replaces that buffer's node with one which points to the perspective's
+*scratch* buffer."
```
zenspider: idgi

perspective.el
conao3: OK, revert this thange.

perspective.el
conao3: This diff is just `M-q` and lisp conventions. Double-space is needed after a period.

perspective.el
conao3: OK, revert this change.

conao3: How about `Cancelled persp-state-save`?

perspective.el
gcv: @conao3: Thank you for looking into this! I very much appreciate the effort you put into improving the docstrings, especially where you added clarifications about parameter behavior.

That said, can we please _not_ use two spaces after a period in perspective source code? The convention you're referring to has not been recommended for over 25 years in the Chicago Manual of Style (https://www.chicagomanualofstyle.org/qanda/data/faq/topics/OneSpaceorTwo/faq0007.html) or the Modern Language Association (https://style.mla.org/number-of-spaces-after-period/), the two most authoritative bodies in American English writing. On a personal note, I have always found two spaces profoundly difficult to look at in monospaced typefaces. It leads to a jagged look, and makes the problem of rivers in text worse (https://en.wikipedia.org/wiki/River_(typography)). `(setq sentence-end-double-space nil)` is one of the earliest entries I put in my Emacs configuration. I know it breaks sentence movement commands, but the tradeoff is worth it to me.

conao3: Long-time no see @gcv! Since this issue ( https://github.com/tumashu/posframe/issues/29 ).
It is very good that an active maintainer can manage this package.
This seems to be different agreed upon, and in this case, the maintainer's opinion is respected.
Close PR.

gcv: I didn't mean for you to close the whole PR, and apologize if my comment was too harsh. I merged in most of your suggestions in commit https://github.com/nex3/perspective-el/commit/599aa7df7805c507cf06eeb1756d6388f1b3b3fd.

conao3: No need to apologize! And it is good you become a collaborator!

PR #108, opened 2019-11-13
Fix a problem with Emacs Childframe

### LINKED ISSUE #79: Incompatible with lsp-ui
Emacs version - 26.1

I use perspective in pair with [lsp-ui](https://github.com/emacs-lsp/lsp-ui).

When [lsp-ui-doc](https://github.com/emacs-lsp/lsp-ui#lsp-ui-doc) feature called (popup with documentation), I got: `error in process filter: persp-add-buffer: Wrong type argument: perspective, nil`.

On MacOS High Sierra in fullscreen mode, desktop workspace with full screen empty Emacs window created after each of error.

On Ubuntu 18.04 (Emacs 25.3.2) I don't know why, but I don't have this error.

I found a topic with similar error and its dated half year ago:
 https://www.reddit.com/r/emacs/comments/93kwtu/how_do_i_debug_a_function_thats_being_called_when/

Ailrun:
Close #79.
Current code does not work when a target frame is a child frame. This patch does not make perspective use the Childframe feature properly, but at least make it work with the feature.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -811,9 +811,11 @@ See also `persp-add-buffer'."
     (when ad-return-value
       (let ((buf (ad-get-arg 0))
             (frame (window-frame ad-return-value)))
-        (when (and buf frame)
-          (with-selected-frame frame
-            (persp-add-buffer buf)))))))
+        (unless (and (version<= "26" emacs-version)
+                     (frame-parent frame))
+          (when (and buf frame)
+            (with-selected-frame frame
+              (persp-add-buffer buf))))))))
 
 (defadvice set-window-buffer (after persp-add-buffer-adv)
   "Add BUFFER to the perspective for window's frame.
@@ -822,9 +824,11 @@ See also `persp-add-buffer'."
   (persp-protect
     (let ((buf (ad-get-arg 1))
           (frame (window-frame (ad-get-arg 0))))
-      (when (and buf frame)
-        (with-selected-frame frame
-          (persp-add-buffer buf))))))
+      (unless (and (version<= "26" emacs-version)
+                   (frame-parent frame))
+        (when (and buf frame)
+          (with-selected-frame frame
+            (persp-add-buffer buf)))))))
 
 (defadvice switch-to-prev-buffer (around persp-ensure-buffer-in-persp)
   "Ensure that the selected buffer is in WINDOW's perspective."
--- a/perspective.el
+++ b/perspective.el
@@ -794,6 +794,11 @@ is non-nil or with prefix arg, don't switch to the new perspective."
         (persp-update-modestring)
       (persp-activate persp))))
 
+(defun persp-child-frame-p (frame)
+  "Check whether FRAME is a child frame or not."
+  (and (version<= "26" emacs-version)
+       (frame-parent frame)))
+
 (defadvice switch-to-buffer (after persp-add-buffer-adv)
   "Add BUFFER to the current perspective.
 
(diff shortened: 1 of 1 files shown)


visortelle:
It works. But still not ideal. On MacOS in Emacs fullscreen mode + [lsp-ui](https://github.com/emacs-lsp/lsp-ui/issues/191), whole screen blinking.

visortelle:
Attached video for the my previous comment. Sorry for google drive.

https://drive.google.com/open?id=12fLZxG62_zqvP45pSgnH4B3q4ocKIYL_

Ailrun:
I don't have a MacOS device, so I have no idea. If you know the reason why, or at least some candidates of the reason, it would be great.

visortelle:
@Ailrun sorry, but I have no idea how to fix it. In any case your patch at least works. :)

Ailrun:
@zenspider any thoughts?

zenspider:
@visortelle normally videos are great at demonstrating how problems occur... but please please please, next time, turn off your theme and increase the font size (or drop full screen). Yes, I have old eyes, but your theme is dark on darker and your video gets compressed so it's just hard.

zenspider:
I'm going slow and haven't had any coffee yet...

@Ailrun can you walk me through this? I don't use lsp or lsp-ui or do much of anything with child-frames... This is happening because lsp-ui is using new features like child frames w/o guarding it? or?

As an aside (I can push this into a code review feedback thingy), the duplication of code makes me think that this should be refactored outright... Both the conditional should be refactored to `frame-parent-p`, but also that this overall action should be refactored to something like `persp-maybe-add-buffer` (or something better named). Or maybe this extra guard logic needs to get pushed down to `persp-add-buffer`?

Ailrun:
@zenspider
> Both the conditional should be refactored to `frame-parent-p`

`frame-parent-p` is not a function, and `frame-ancestor-p` requires 2 frames, one for child and one for parent. ([Ref](https://www.gnu.org/software/emacs/manual/html_node/elisp/Child-Frames.html))

> This is happening because lsp-ui is using new features like child frames w/o guarding it? or?

This happens, I guess, because `display-buffer-in-child-frame` creates a (child) frame, but it does not call `after-make-frame-functions` on the frame. Therefore, the childframe does not have persp properties, and returns `nil` when you access one.

> Or maybe this extra guard logic needs to get pushed down to `persp-add-buffer`?

That's what I originally thought, but I wasn't sure which frame should I access to determine whether target frame of buffer addition is a child or not. Could you guide me which one it is?

zenspider:
> frame-parent-p is not a function

Sorry, I wasn't clear... I was saying that `frame-parent-p` should be written to refactor the condition clause to make the intent more clear.

> Could you guide me which one it is?

No, I probably couldn't. 😛 If it doesn't fit, it doesn't fit. That was an un-caffeinated guess.

Ailrun:
@zenspider I meant, `persp-add-buffer` does not have any `frame` as an argument or anything, but emacs has current (selected) frame, visible frames, default-minibuffer-frame, ... etc.. I guess one of  them should be the target frame, but I'm not sure which one it is.

gcv:
@Ailrun: Your change looks fine, and may in fact be a very good one. I'm still thinking about it. Before we merge it in, though, could you please try using `lsp-ui` with the latest master and report back? Reason being: we recently merged in https://github.com/nex3/perspective-el/commit/29f4c5cad502ec73d4c62199a9d940b318351dd3, which is another attempt to tackle the problem of uninitialized frames.

@visortelle: I watched your video. Are you using the official NextStep or Mitsuharu Yamamoto's Mac Emacs port? `window-system` will display `ns` for NextStep, `mac` for Mac. I have had problems with child frames breaking in full-screen mode due to problems with the NextStep port, but they work fine on the Mac. At least, I don't see the problem in your video. I use child frames with `ivy-posframe`, though, and not `lsp-ui`.

Ailrun:
OK, I will check it. Thanks for the recommendation.

Ailrun:
@gcv I think it works! (at least it works for me)
@visortelle Could you also check whether it works?

visortelle:
@Ailrun: I am in progress :)

visortelle:
@Ailrun, @gcv: Mitsuharu Yamamoto's Mac Emacs port works as expected, without any visual artifacts.

Ailrun:
@visortelle @zenspider @gcv OK, I think it's OK to close this PR. Thank you for your coorperation!

PR #109, opened 2019-11-24
Removed warnings that come up from byte compiling.

zenspider:
This doesn't deal with lexical-let + cl.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -191,6 +191,9 @@ Run with the activated perspective active.")
 (defvar persp-mode-map (make-sparse-keymap)
   "Keymap for perspective-mode.")
 
+(defvar perspective-map nil
+  "Sub-keymap for perspective-mode")
+
 (define-prefix-command 'perspective-map)
 (define-key persp-mode-map persp-mode-prefix-key 'perspective-map)
 
--- a/test/test-perspective.el
+++ b/test/test-perspective.el
@@ -31,12 +31,12 @@ tell-tale leading '*' characters)."
 (defun persp-test-buffer-list-all ()
   "Return the list of buffers in Emacs (in all
 perspectives), filtering out temporaries."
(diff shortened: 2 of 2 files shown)


zenspider:
@gcv I just rebased against latest master and got a new one... Does this seem reasonable?

```diff
-                (state-complete (read-from-whole-string
-                                 (with-temp-buffer
-                                   (insert-file-contents file)
-                                   (buffer-string)))))
+                (state-complete (with-temp-buffer
+                                  (insert-file-contents-literally path)
+                                  (buffer-string))))
```

I don't understand the use of `read-from-whole-string` (now `thing-at-point--read-from-whole-string`) in this context... So my change might not cover what you want.

gcv:
The intent is to call `read` on the persisted state file. I switched `read-from-whole-string` because it flags an error if there's more stuff left after a successful `read` call. The difference in behavior between `(read "(list 1 2) (list 3 4)")` and `(read-from-whole-string "(list 1 2) (list 3 4)")` is that the first just returns `(1 2)` and the second errors out. This was helpful in flagging situations when `persp-state-save` just appended state to an existing file instead of overwriting it (or prompting).

There should now be adequate logic to prevent that from happening, and I haven't seen this error pop up in many months. The choices are:

1. Switch to using `thing-at-point--read-from-whole-string`.
2. Copy the implementation of `thing-at-point--read-from-whole-string` into perspective.
3. Switch to vanilla `read` and hope all the overwriting logic works.

zenspider:
I went with `thing-at-point--read-from-whole-string` in the end, but I think I'd prefer `read` since the file format is just one sexp.

zenspider:
I got it fixed up for 24.4. We can switch to read in a separate PR. I think these two PRs can go in. I'll file an issue for the switch.

PR #112, opened 2019-11-26
Attempt to switch to lexical scope

zenspider:


zenspider:
I just put in a fix for 24.4 which didn't have the `thing-at-pt` function so I aliased it. The test was an easy conversion. Looks like this might be good to go?

zenspider:
I've been working with these for a week now w/o problems... I'm gonna merge both in to master (probably locally to avoid PR on PR issues I've had in the past)

gcv:
Should we delete the merged branches `lexical-scope` and `byte-compile-cleanup`?

zenspider:
I usually delete remotely via `git remote update --prune`.

PR #116, opened 2019-12-02
Remove dead code and refactor

zenspider:
MIGHT make more sense to do this commit-by-commit.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -123,6 +123,10 @@ After BODY is evaluated, frame parameters are reset to their original values."
   "Return a list of all buffers in the current perspective."
   `(persp-buffers (persp-curr)))
 
+(defun persp-is-current-buffer (buf)
+  "Return T if BUF is in the current perspective."
+  (memq buf (persp-current-buffers)))
+
 (defalias 'persp-killed-p 'persp-killed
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -139,6 +119,37 @@ After BODY is evaluated, frame parameters are reset to their original values."
   (window-configuration (current-window-configuration))
   (point-marker (point-marker)))
 
+(defmacro persp-current-buffers ()
```
gcv: Just curious, why make this a macro?

zenspider: Because of `setf` below... If it is a function, `setf` doesn't work (presumably because you get a copy?).

gcv: Oh, I didn't notice the use of `setf`. It doesn't work with a function because `setf` does magic things and expects a `place` list (https://www.gnu.org/software/emacs/manual/html_node/elisp/Setting-Generalized-Variables.html).

zenspider:
Are you OK with the names I wound up with? I'm not thrilled...

I kinda want to do a sweep to normalize the names.

gcv:
What naming changes do you have in mind? I’m fine with the ones you chose. Naming things is hard. :)

PR #117, opened 2019-12-05
added check for when frame-parameter returns a string

Gopiandcode:
I haven't fully grokked the internals of perspective yet to understand whether this is a complete fix, but it seems to handle the issue I was facing of repeatedly receiving error messages of the form:
```
Wrong type argument: perspective, "Unprintable entity"
```
When using perspective in combination with certain packages such as desktop-mode (I think?), for some reason `(frame-parameter frame 'persp--curr)` returns a string "Unprintable entity" rather than nil, causing persp-curr to return a string rather than a perspective-object.

This seems to be the cause of this issue in spacemacs: https://github.com/syl20bnr/spacemacs/issues/2626.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -242,8 +242,14 @@ perspective-local values."
 FRAME defaults to the currently selected frame."
   ;; XXX: This must return a non-nil value to avoid breaking frames initialized
   ;; with after-make-frame-functions bound to nil.
-  (or (frame-parameter frame 'persp--curr)
-      (make-persp-internal)))
+  (let ((frame-persp (frame-parameter frame 'persp--curr)))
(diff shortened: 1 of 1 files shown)


Gopiandcode:
On closer inspection, I don't think this fixes the issue as it seems like the origin of this string value for the frame parameter is  not from from just `persp-curr`.

If anyone has any idea as to where this "Unprintable" value might arise from, I would appreciate it, but for now I'll close this PR.

gcv:
The spacemacs ticket you linked to says desktop-mode corrupts frame parameters if they contain window or buffer state. That’s pretty much the end of that.

Do you really need desktop-mode? Can you use the new(-ish) persp-state feature to save perspectives instead?

PR #119, opened 2019-12-15
Remove buffer-name-history from perspective state.

### LINKED ISSUE #115: Internal: Investigate the use of buffer-history and buffer-name-history
It's tracked in the `perspectives-hash`, gets huge and stale, and is copied to every perspective.

It seems to only be relevant to the vanilla (i.e., not ido or any other replacement) implementation of `switch-to-buffer` (https://www.emacswiki.org/emacs/MinibufferHistory).

I think the intent was to keep perspective-specific buffer history around to help narrow down search results (i.e., `C-x b` then `C-r`), but right now, it brings up _all_ buffer history.

gcv:
It was being tracked in `perspectives-hash`, and was being copied to every
perspective on activation. So each perspective ended up with the same large copy
of the buffer history list.

It seems to only be relevant to the vanilla (i.e., not ido or any other replacement) implementation of `switch-to-buffer` (https://www.emacswiki.org/emacs/MinibufferHistory). I think the intent was to keep perspective-specific buffer history around to help narrow down search results (i.e., `C-x b` then `C-r`), but right now, it does no perspective-specific history filtering at all.

With this patch, the perspective hash becomes much smaller and cleaner, and perspective activation should be just a little bit faster since there is no need to copy a very large history list around every time.

Am I missing something here? @zenspider, @dzop, @conao3 — your opinions, please?

If this makes sense, it would resolve #115.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -115,7 +115,6 @@ After BODY is evaluated, frame parameters are reset to their original values."
   name buffers killed local-variables
   (last-switch-time (current-time))
   (created-time (current-time))
-  (buffer-history buffer-name-history)
   (window-configuration (current-window-configuration))
   (point-marker (point-marker)))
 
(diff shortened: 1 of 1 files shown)


conao3:
I'm nothing on it.

gcv:
I've been using this patch for the last few weeks, and it works well. I'll merge it in soon if I don't hear any objections.

PR #120, opened 2020-01-05
feat(perspective.el): args are optional for persp-state-load or save

nickgarber:
- adjusted default arguments for function persp-state-load
 - adjusted default arguments for function persp-state-save

These were the changes I needed to make to get perspective to automatically save and restore when Emacs stops and starts. Just wanted to make the changes available.

Here's my use-package for context, using Emacs version 26.3 on Ubuntu 19.10
```
    (use-package perspective
      :demand t
      :commands (persp-state-load persp-state-save)
      :hook
      (after-init . persp-state-load) ;; load perspectives on initialization
      (kill-emacs . persp-state-save) ;; save perspectives on exit
      :config
      (persp-mode t))
```

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1166,13 +1166,20 @@ visible in a perspective as windows, they will be saved as
 '*scratch* (persp)' buffers."
   (interactive (list
                 (read-file-name "Save perspective state to file: "
-                                persp-state-default-file
-                                persp-state-default-file)
-                t))
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-directory-p (file-name-directory persp-state-default-file)))
+                                    (file-name-directory persp-state-default-file)
+                                  "")
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-readable-p persp-state-default-file))
+                                    (file-name-nondirectory persp-state-default-file)
+                                  ""))))
   (unless persp-mode
     (message "persp-mode not enabled, nothing to save")
     (cl-return-from persp-state-save))
-  (let ((target-file (if (and file (not (string-equal "" file)))
+  (let ((target-file (if (and (bound-and-true-p file) (not (string-equal "" file)))
                          ;; file provided as argument, just use it
                          (expand-file-name file)
                        ;; no file provided as argument
@@ -1183,7 +1190,7 @@ visible in a perspective as windows, they will be saved as
                            nil
                          ;; in non-interactive call mode, we want to fall back to
                          ;; the default, but only if it is set
-                         (if (and persp-state-default-file
+                         (if (and (bound-and-true-p persp-state-default-file)
                                   (not (string-equal "" persp-state-default-file)))
                              (expand-file-name persp-state-default-file)
                            nil)))))
@@ -1216,7 +1223,7 @@ visible in a perspective as windows, they will be saved as
     (run-hooks 'persp-state-after-save-hook)))
 
 ;;;###autoload
-(defun persp-state-load (file)
+(defun persp-state-load (&optional file)
   "Restore the perspective state saved in FILE.
 
 FILE defaults to the value of persp-state-default-file if it is
@@ -1227,10 +1234,23 @@ Each perspective's buffer list and window layout are also
 restored."
   (interactive (list
                 (read-file-name "Restore perspective state from file: "
-                                persp-state-default-file
-                                persp-state-default-file)))
-  (unless (file-exists-p file)
-    (error "File not found"))
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-directory-p (file-name-directory persp-state-default-file)))
+                                    (file-name-directory persp-state-default-file)
+                                  user-emacs-directory)
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-readable-p persp-state-default-file))
+                                    (file-name-nondirectory persp-state-default-file)
+                                  "perspective.persist.el"))))
+  ;;
+  (or (and
+       (not (bound-and-true-p file))
+       (bound-and-true-p persp-state-default-file)
+       (file-readable-p persp-state-default-file)
+       (setq file (expand-file-name (file-name-nondirectory persp-state-default-file) (file-name-directory persp-state-default-file))))
+      (error "unable to use variable persp-state-default-file"))
   (persp-mode 1)
   ;; before hook
   (run-hooks 'persp-state-before-load-hook)


perspective.el
```diff
@@ -1166,13 +1166,20 @@ visible in a perspective as windows, they will be saved as
...
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-readable-p persp-state-default-file))
+                                    (file-name-nondirectory persp-state-default-file)
+                                  ""))))
   (unless persp-mode
     (message "persp-mode not enabled, nothing to save")
     (cl-return-from persp-state-save))
-  (let ((target-file (if (and file (not (string-equal "" file)))
+  (let ((target-file (if (and (bound-and-true-p file) (not (string-equal "" file)))
```
gcv: I think this is the reason why this patch breaks the test suite. This is not correct use of `bound-and-true-p`: it will return `nil` when called interactively because `file` is in lexical scope. The documentation for `boundp` states:

> Note that if ‘lexical-binding’ is in effect, this refers to the
global value outside of any lexical scope.

We now use `lexical-binding` in Perspective.

perspective.el
```diff
@@ -1227,10 +1234,23 @@ Each perspective's buffer list and window layout are also
...
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-directory-p (file-name-directory persp-state-default-file)))
+                                    (file-name-directory persp-state-default-file)
+                                  user-emacs-directory)
+                                (if (and
+                                     (bound-and-true-p persp-state-default-file)
+                                     (file-readable-p persp-state-default-file))
+                                    (file-name-nondirectory persp-state-default-file)
+                                  "perspective.persist.el"))))
```
gcv: Same comments on this as for `persp-state-save`. The call for `read-file-name` should not change, though you may want to add the `interactive?` parameter and corresponding `t` as for (unpatched) `persp-state-save`.

perspective.el
```diff
@@ -1227,10 +1234,23 @@ Each perspective's buffer list and window layout are also
...
+  ;;
+  (or (and
+       (not (bound-and-true-p file))
+       (bound-and-true-p persp-state-default-file)
+       (file-readable-p persp-state-default-file)
+       (setq file (expand-file-name (file-name-nondirectory persp-state-default-file) (file-name-directory persp-state-default-file))))
+      (error "unable to use variable persp-state-default-file"))
```
gcv: - Use of short-circuit boolean logic to modify a variable with `or ... setq` is pretty obscure in Lisp code. If you need to mutate a variable, please use an explicit `if` or `when`.
- `(bound-and-true-p file)` does not work for lexically bound `file` as discussed above, just use plain `file` to check as it will always be bound and set to `nil` by default.
- I'm pretty sure the right thing here is to just copy the `(let ((target-file ...` binding from `persp-state-save` and use it.

I played with this implementation, and it seems to work correctly:

```
(defun persp-state-load (&optional file interactive?)
  "Restore the perspective state saved in FILE.

FILE defaults to the value of persp-state-default-file if it is
set.

Frames are restored, along with each frame's perspective list.
Each perspective's buffer list and window layout are also
restored."
  (interactive (list
                (read-file-name "Restore perspective state from file: "
                                persp-state-default-file
                                persp-state-default-file)))
  (let ((target-file (if (and file (not (string-equal "" file)))
                         ;; file provided as argument, just use it
                         (expand-file-name file)
                       ;; no file provided as argument
                       (if interactive?
                           ;; return nil in interactive call mode, since
                           ;; read-file-name should have provided a reasonable
                           ;; default
                           nil
                         ;; in non-interactive call mode, we want to fall back to
                         ;; the default, but only if it is set
                         (if (and persp-state-default-file
                                  (not (string-equal "" persp-state-default-file)))
                             (expand-file-name persp-state-default-file)
                           nil)))))
    (unless (file-exists-p target-file)
      (error "File not found"))
    (persp-mode 1)
    ;; before hook
    (run-hooks 'persp-state-before-load-hook)
    ;; actually load
    (let ((tmp-persp-name (format "%04x%04x" (random (expt 16 4)) (random (expt 16 4))))
          (frame-count 0)
          (state-complete (read
                           (with-temp-buffer
                             (insert-file-contents-literally target-file)
                             (buffer-string)))))
      ;; open all files in a temporary perspective to avoid polluting "main"
      (persp-switch tmp-persp-name)
      (cl-loop for file in (persp--state-complete-files state-complete) do
               (when (file-exists-p file)
                 (find-file file)))
      ;; iterate over the frames
      (cl-loop for frame in (persp--state-complete-frames state-complete) do
               (cl-incf frame-count)
               (when (> frame-count 1)
                 (make-frame-command))
               ;; XXX: The condition in binding frame-persp-table and
               ;; frame-persp-order ensures backwards compatibility with the early
               ;; pre-release format of persp-state files: the frame may be just a
               ;; hash table (old version), or it may be an instance of
               ;; persp--state-frame (released version). The special case can be
               ;; removed in the future, as there should be very few or no files
               ;; left in the old format.
               (let* ((frame-persp-table (if (hash-table-p frame)
                                             frame
                                           (persp--state-frame-persps frame)))
                      (frame-persp-order (if (hash-table-p frame)
                                             (hash-table-keys frame)
                                           (reverse (persp--state-frame-order frame)))))
                 ;; iterate over the perspectives in the frame in the appropriate order
                 (cl-loop for persp in frame-persp-order do
                          (let ((state-single (gethash persp frame-persp-table)))
                            (persp-switch persp)
                            (cl-loop for buffer in (persp--state-single-buffers state-single) do
                                     (persp-add-buffer buffer))
                            ;; XXX: split-window-horizontally is necessary for
                            ;; window-state-put to succeed? Something goes haywire with root
                            ;; windows without it.
                            (split-window-horizontally)
                            (window-state-put (persp--state-single-windows state-single)
                                              (frame-root-window (selected-frame))
                                              'safe)))))
      ;; cleanup
      (persp-kill tmp-persp-name))
    ;; after hook
    (run-hooks 'persp-state-after-load-hook)))
```

The binding of `target-file` is identical in `persp-state-load` and `persp-state-save`, and should be refactored accordingly.

nickgarber:
I'm glad to know you're interested in adopting parts of this. Please allow me a bit of time to fully comprehend all your comments? I appreciate your insights and hope to make the most of them.

gcv:
Of course. Your idea is a good one, since I'm sure other people would want to auto-load perspectives on Emacs start. Feel free to ask questions if any of my comments don't make sense.

keithmantell:
Just to say that the load with the default file variable would be a great addition!

nickgarber:
I'm closing this PR but feel free to re-open and re-own if anyone else would like to pursue it. Thanks!

PR #121, opened 2020-01-24
Fix byte-compile error

aaronjensen:
Eager macro-expansion failure: (void-variable ido-ignore-buffers)

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -34,7 +34,7 @@
 (require 'bs)                           ; buffer switcher
 
 (defvar ido-temp-list)
-(defvar ido-ignore-buffers)
+(defvar ido-ignore-buffers nil)
 
 ;;; Code:
 


gcv:
I tried compiling the package using melpa's standalone package sandbox machinery, and I still get an error with this change. `perspective.el:1029:1:Error: Symbol’s value as variable is void: ido-ignore-buffers`.

I don't understand this error at all.

gcv:
It seems to go away with

```
(eval-when-compile
  (require 'ido)
  (require 'ibuffer))
```

aaronjensen:
Interesting. I wonder why I was able to compile. Probably a dirty environment.

I think the issue is probably this form:

```elisp
(rx (eval (append (list 'or)
  (mapcar (lambda (rx) `(regexp ,rx))
  ido-ignore-buffers))))
```

That actually expands at compile time and the entire thing evals, including `ido-ignore-buffers` I believe, which probably isn't actually what you want. I don't know enough to say how to fix it, but hopefully that's helpful.

aaronjensen:
Try changing `rx` to `rx-to-string`? https://emacs.stackexchange.com/questions/2288/how-do-i-create-a-dynamic-regexp-with-rx 

You probably need to drop the eval as well.

gcv:
Thanks! You gave me the clues I need to (I think) fix the problem. Eager macro-expansion in Elisp is a helluva footgun.

gcv:
Common Lisp goes to considerable lengths to avoid this kind of difference in behavior in compiled and interpreted code.

aaronjensen:
No problem, though did my PR help anything? I think it's a workaround in some scenarios that makes it ultimately behave incorrectly (because `ido-ignore-buffers` will always be nil when byte compiled), but I assume you have another commit coming after you validate?

gcv:
Yup, working on a patch right now that (among other things) uses `rx-to-string` instead of `rx`.

aaronjensen:
That's great, thank you. And thank you for breathing new life into this project, I appreciate it.

PR #125, opened 2020-01-30
Preselect the other buffer in persp-counsel-switch-buffer.

dme:
When switching buffers, the default selection should be the other
buffer rather than the current buffer.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1080,32 +1080,48 @@ PERSP-SET-IDO-BUFFERS)."
         (ibuffer))
     (ibuffer)))
 
-;; Buffer switching integration: Ivy / Counsel.
-;;;###autoload
-(defun persp-counsel-switch-buffer (arg)
-  "Like COUNSEL-SWITCH-BUFFER, but Perspective-aware.
-With a prefix arg, show buffers in all perspectives."
-  (interactive "P")
-  (unless (and (featurep 'ivy) (featurep 'counsel))
-    (error "Ivy or Counsel not loaded"))
+(defun persp--switch-buffer-ivy-counsel-helper (arg ivy-params fallback)
+  (unless (featurep 'ivy)
+    (error "Ivy not loaded"))
   (defvar ivy-switch-buffer-map)
   (declare-function ivy-read "ivy.el")
   (declare-function ivy--switch-buffer-matcher "ivy.el")
   (declare-function ivy--switch-buffer-action "ivy.el")
+  (if (and persp-mode (null arg))
+      (apply #'ivy-read
+             (append
+              (list
+               (format "Switch to buffer (%s): " (persp-current-name))
+               (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
+               :preselect (buffer-name (persp-other-buffer (current-buffer)))
+               :keymap ivy-switch-buffer-map
+               :action #'ivy--switch-buffer-action
+               :matcher #'ivy--switch-buffer-matcher)
+              ivy-params))
+    (funcall fallback)))
+
+;; Buffer switching integration: Ivy.
+;;;###autoload
+(defun persp-ivy-switch-buffer (arg)
+  "A version of `ivy-switch-buffer' which respects perspectives."
+  (interactive "P")
+  (persp--switch-buffer-ivy-counsel-helper arg nil #'ivy-switch-buffer))
+
+;; Buffer switching integration: Counsel.
+;;;###autoload
+(defun persp-counsel-switch-buffer (arg)
+  "A version of `counsel-switch-buffer' which respects perspectives."
+  (interactive "P")
+  (unless (featurep 'counsel)
+    (error "Counsel not loaded"))
   (declare-function counsel-switch-buffer "counsel.el")
   (declare-function counsel--switch-buffer-unwind "counsel.el")
   (declare-function counsel--switch-buffer-update-fn "counsel.el")
-  (if (and persp-mode (null arg))
-      (ivy-read (format "Switch to buffer (%s): " (persp-current-name))
-                (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
-                :preselect (buffer-name (persp-other-buffer (current-buffer)))
-                :keymap ivy-switch-buffer-map
-                :action #'ivy--switch-buffer-action
-                :matcher #'ivy--switch-buffer-matcher
-                :caller #'counsel-switch-buffer
-                :unwind #'counsel--switch-buffer-unwind
-                :update-fn #'counsel--switch-buffer-update-fn)
-    (counsel-switch-buffer)))
+  (persp--switch-buffer-ivy-counsel-helper arg
+                                           (list :caller #'counsel-switch-buffer
+                                                 :unwind #'counsel--switch-buffer-unwind
+                                                 :update-fn #'counsel--switch-buffer-update-fn)
+                                           #'counsel-switch-buffer))
 
 ;; Symbols namespaced by persp--state (internal) and persp-state (user
 ;; functions) provide functionality which allows saving perspective state on
--- a/perspective.el
+++ b/perspective.el
@@ -1021,6 +1021,17 @@ perspective beginning with the given letter."
   (setq persp-show-modestring t)
   (persp-update-modestring))
 
+(defun persp-other-buffer (&optional buffer visible-ok frame)
+  "A version of `other-buffer' which respects perspectives."
+  (let ((other (other-buffer buffer visible-ok frame)))
+    (if (member other (persp-current-buffers))
+	other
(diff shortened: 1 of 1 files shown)


gcv:
Sounds good overall.

I think that in some cases (which should be somewhat rare after #95 but may come up) `other-buffer` may refer to a buffer outside the current perspective. Please put in a check using `persp-is-current-buffer`, and fall back to preselecting the current buffer in that case.

On a related note, I've been eyeing the `:update-fn` setting in the call to `ivy-read`, and realized I don't like it. The jump to previewing the other buffer in the current window is too jarring for me. That seems that is the only substantive difference between `ivy-switch-buffer` (which does not switch as the selected item changes) and `counsel-switch-buffer` (which does). I'm not sure what to do here. Maybe copy-paste (yuck) `persp-counsel-switch-buffer` into `persp-ivy-switch-buffer`, remove that line, document the distinction, and let people pick which implementation they prefer?

dme:
> I think that in some cases (which should be somewhat rare after #95
> but may come up) `other-buffer` may refer to a buffer outside the
> current perspective. Please put in a check using
> `persp-is-current-buffer`, and fall back to preselecting the current
> buffer in that case.

I added `persp-other-buffer' to try and avoid this problem. It's force
pushed to my own repo and I did the pull request thing, but I'm unsure
what github will do in that case.

> On a related note, I've been eyeing the `:update-fn` setting in the
> call to `ivy-read`, and realized I don't like it. The jump to
> previewing the other buffer in the current window is too jarring for
> me. That seems that is the only substantive difference between
> `ivy-switch-buffer` (which does not switch as the selected item
> changes) and `counsel-switch-buffer` (which does). I'm not sure what
> to do here. Maybe copy-paste (yuck) `persp-counsel-switch-buffer` into
> `persp-ivy-switch-buffer`, remove that line, document the distinction,
> and let people pick which implementation they prefer?

Another changeset is included that tries to address this. It's not well
tested - more for an idea of approach.

dme.
-- 
When I grow up I'll be stable.

perspective.el
```diff
@@ -28,6 +28,7 @@
 ;; available by default.
 
 (require 'cl-lib)
+(require 'cl-macs)
```
gcv: This require is not necessary for anything, including `do`.

perspective.el
```diff
@@ -1021,6 +1022,13 @@ perspective beginning with the given letter."
   (setq persp-show-modestring t)
   (persp-update-modestring))
 
+(defun persp-other-buffer (&optional buffer visible-ok frame)
+  "A version of `other-buffer' which respects perspectives."
+  (let ((candidates (persp-current-buffers)))
+    (do ((buf (other-buffer buffer visible-ok frame)
```
gcv: Minor problem: this should be `cl-do` for consistency with the rest of the code and to keep the byte-compiler happy. We had to standardize use of `loop` to `cl-loop` in the past for these reasons.

Bigger problem: this code started an infinite loop when I called it in a new perspective.

perspective.el
```diff
@@ -1069,32 +1077,47 @@ PERSP-SET-IDO-BUFFERS)."
...
-    (error "Ivy or Counsel not loaded"))
+(defun persp--switch-buffer-1 (arg ivy-params fallback)
+  (unless (featurep 'ivy)
+    (error "Ivy not loaded"))
   (defvar ivy-switch-buffer-map)
   (declare-function ivy-read "ivy.el")
   (declare-function ivy--switch-buffer-matcher "ivy.el")
   (declare-function ivy--switch-buffer-action "ivy.el")
+  (if (and persp-mode (null arg))
+      (apply 'ivy-read
```
gcv: This should be `#'ivy-read` for the byte-compiler.

perspective.el
```diff
@@ -1069,32 +1077,47 @@ PERSP-SET-IDO-BUFFERS)."
...
 
-;; Buffer switching integration: Ivy / Counsel.
-;;;###autoload
-(defun persp-counsel-switch-buffer (arg)
-  "Like COUNSEL-SWITCH-BUFFER, but Perspective-aware.
-With a prefix arg, show buffers in all perspectives."
-  (interactive "P")
-  (unless (and (featurep 'ivy) (featurep 'counsel))
-    (error "Ivy or Counsel not loaded"))
+(defun persp--switch-buffer-1 (arg ivy-params fallback)
```
gcv: Let's call this function `persp--switch-buffer-ivy-counsel-helper`, since it's so specific to these packages.

perspective.el
```diff
@@ -1069,32 +1077,47 @@ PERSP-SET-IDO-BUFFERS)."
...
+               :matcher #'ivy--switch-buffer-matcher)
+              ivy-params))
+    (funcall fallback)))
+
+;; Buffer switching integration: Ivy.
+;;;###autoload
+(defun persp-ivy-switch-buffer (arg)
+  "A version of `ivy-switch-buffer' which respects perspectives."
+  (interactive "P")
+  (persp--switch-buffer-1 arg nil 'ivy-switch-buffer))
```
gcv: This should be `#'ivy-switch-buffer` for the byte-compiler.

perspective.el
```diff
@@ -1069,32 +1077,47 @@ PERSP-SET-IDO-BUFFERS)."
...
-                :action #'ivy--switch-buffer-action
-                :matcher #'ivy--switch-buffer-matcher
-                :caller #'counsel-switch-buffer
-                :unwind #'counsel--switch-buffer-unwind
-                :update-fn #'counsel--switch-buffer-update-fn)
-    (counsel-switch-buffer)))
+  (persp--switch-buffer-1 arg (list :caller #'counsel-switch-buffer
+                                    :unwind #'counsel--switch-buffer-unwind
+                                    :update-fn #'counsel--switch-buffer-update-fn)
+                          'counsel-switch-buffer))
```
gcv: This should be `#'counsel-switch-buffer` for the byte-compiler.

dme:
On Friday, 2020-01-31 at 11:35:00 -08, gcv wrote:

> I wonder if it would be easier to forget getting the correct
> `other-buffer` in cases when it's not immediately correct and just
> fall back to the current buffer in cases when `other-buffer` is not in
> the current perspective.

Currently I don't understand what causes it to not be immediately
correct (I haven't been able to provoke that), so I'm unsure whether
this is likely to be tolerable. Certainly the current behaviour (where
the current buffer is the default) is horrible - it breaks years of
muscle memory.

> +(require 'cl-macs)
>
> This require is not necessary for anything, including `do`.

Okay.

> +    (do ((buf (other-buffer buffer visible-ok frame)
>
> Minor problem: this should be `cl-do` for consistency with the rest of
> the code and to keep the byte-compiler happy. We had to standardize
> use of `loop` to `cl-loop` in the past for these reasons.

Okay.

> Bigger problem: this code started an infinite loop when I called it in
> a new perspective.

I will write some tests.

> +      (apply 'ivy-read
>
> This should be `#'ivy-read` for the byte-compiler.

Okay, for all of these.

> -(defun persp-counsel-switch-buffer (arg)
>
> Let's call this function `persp--switch-buffer-ivy-counsel-helper`,
> since it's so specific to these packages.

Okay.

dme.
-- 
But they'll laugh at you in Jackson, and I'll be dancin' on a Pony Keg.

gcv:
To test Perspective in isolation, I follow the MELPA procedure: https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#test-your-recipe

The short version of the instructions is: clone the melpa repo, and then change the perspective recipe to point to your local perspective repo:

```
(perspective :fetcher git :url "/home/you/code/perspective-el")
```

Then, from the melpa repo, run:

```
EMACS_COMMAND=/path/to/emacs make recipes/perspective
```

Then install Ivy and Counsel in the sandbox (the sandbox which will be inside your melpa checkout):

```
EMACS_COMMAND=/path/to/emacs make sandbox INSTALL=ivy
# exit emacs after this step
EMACS_COMMAND=/path/to/emacs make sandbox INSTALL=counsel
# exit emacs after this step
```

Finally install perspective:

```
EMACS_COMMAND=/path/to/emacs make sandbox INSTALL=perspective
```

This step should produce no byte-compiler errors or warnings. You'll see why you need to use `cl-do` at this point.

Hmmm. Since I went to the effort of writing all this up, I'd better extract it into a `CONTRIBUTING.org` file. 🙃

To get the infinite loop to occur, the following set of steps do it right after the sandbox Emacs starts:

```
1. M-: (require 'ivy)
2. M-: (persp-mode)
3. M-x persp-switch, call it "one"
4. M-x persp-ivy-switch-buffer
```

After making changes to perspective, you need to commit them locally to perspective (or melpa's recipe builder won't pick them up), rerun `make recipes/perspective`, delete the perspective directory from the sandbox, and rerun `make sandbox INSTALL=perspective`.

> Certainly the current behaviour (where the current buffer is the default) is horrible - it breaks years of muscle memory.

I get it. We'll figure something out.

PR #135, opened 2020-11-06
Fix error symbol’s value as variable is void: union

sillygod:
As a title mention, I encounter this error when I call =persp-import=. I think there is no =union= in emacs lisp but it exists in the common lisp so it should be called =cl-union= instead in emacs-lisp.


gcv:
Great catch! Thank you for contributing to Perspective.

PR #136, opened 2020-11-13
Fix: persp-switch-buffer use ivy-switch-buffer as caller

gcentauri:
Using counsel-switch-buffer was causing issues with the
all-the-icons-ivy integration, as well as preventing all of the ivy
actions from showing up in the hydra menu. for example, killing marked
buffers is not present in the actions if counsel-switch-buffer is
used, but it does appear when the caller is ivy-switch-buffer

this may cause issues for people that have further customization on
the counsel-switch-buffer caller, but from what i can tell, the
counsel-switch-buffer function doesn't even pass itself down as the
caller in its implementation.

### before

image

No icons appear for the buffers, and you can see `[1/5]` actions

### after

image

Icons appear and you can see `[1/8]` actions

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1260,6 +1260,7 @@ PERSP-SET-IDO-BUFFERS)."
                (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
                :preselect (buffer-name (persp-other-buffer (current-buffer)))
                :keymap ivy-switch-buffer-map
+               :caller #'ivy-switch-buffer
                :action #'ivy--switch-buffer-action
                :matcher #'ivy--switch-buffer-matcher)
               ivy-params))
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -1260,6 +1260,7 @@ PERSP-SET-IDO-BUFFERS)."
                (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
                :preselect (buffer-name (persp-other-buffer (current-buffer)))
                :keymap ivy-switch-buffer-map
+               :caller #'ivy-switch-buffer
```
gcentauri, line 1263: this allows it to work for both counsel/ivy switch buffer functions

gcv:
Thank you!

PR #140, opened 2020-12-02
Don't invoke persp-switch-hook on norecord

cjlucas:
`persp-before-switch-hook` and `persp-switch-hook` are being invoked in unexpected cases. Specifically, when `persp-new` is called, the aforementioned hooks are invoked twice due to the use of `persp-switch` in the `with-perspective` macro. Ideally, these hooks should only be invoked as a result of a user action (either switching perspectives interactively or via a bridge like persp-projectile).


gcv:
Looks good to me. Thank you for your contribution!

Issue #142, opened 2020-12-15
persp-ivy-switch-buffer doesn't keep the buffer list order


QiangF:
ivy-switch-buffer by default orders the buffer list by access time, which is broken in the persp command.

PR #143, opened 2020-12-15
sort buffer list by display time in persp-ivy-switch-buffer

QiangF:
fix https://github.com/nex3/perspective-el/issues/142

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1259,7 +1259,11 @@ PERSP-SET-IDO-BUFFERS)."
              (append
               (list
                (format "Switch to buffer (%s): " (persp-current-name))
-               (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
+               (cl-remove-if #'null (mapcar #'buffer-name
+                                            ;; buffer-list is ordered by access time
+                                            ;; seq-intersection keeps the order
+                                            (seq-intersection (buffer-list)
+                                                              (persp-current-buffers))))
                :preselect (buffer-name (persp-other-buffer (current-buffer)))
                :keymap ivy-switch-buffer-map
                :caller #'ivy-switch-buffer


perspective.el
```diff
@@ -30,6 +30,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'dash)
```
gcv, line 33: I'm trying to keep Perspective dependency-free. Even dash is too much here, and its functionality should be reimplemented.

perspective.el
```diff
@@ -1259,7 +1260,8 @@ PERSP-SET-IDO-BUFFERS)."
              (append
               (list
                (format "Switch to buffer (%s): " (persp-current-name))
-               (cl-remove-if #'null (mapcar #'buffer-name (persp-current-buffers)))
+               (cl-remove-if #'null (mapcar #'buffer-name (-intersection (buffer-list)
```
gcv, line 1263: I'm really confused how `-intersection` helps sort by access time here. Please clarify.

gcv (changes requested):

perspective.el
QiangF: By default (buffer-list) is sorted by access time and the order is kept in the -intersection operation.

perspective.el
QiangF: I think we can use seq-intersection, or why not keep persp-current-buffers sorted by time in the first place.

gcv: Please switch to `seq-intersection` and verify that it still works. If it breaks CI for older version of Emacs, we can maybe hack around the problem.

perspective.el
gcv: Very clever! Please add a code comment with this explanation.

gcv:
Thanks!

Issue #145, opened 2021-01-19
change key binding (but not the prefix)


dams:
Hi,

Thank you for this awesome package. I'm using it with a custom prefix key (so far so good) but I'd like to change the key bindings for "next" and "prev".
I'd like to change it from "n" to "C-n"

I've seen it's definied here: https://github.com/nex3/perspective-el/blob/master/perspective.el#L288
But it doesn't seem to be customizable. Is there any way I can change these bindings without changing the package code ?

Thanks

gcv:
You can always override the binding after loading the package. With `use-package`, something like this should work:
```elisp
(use-package perspective
  ;; ...
  :bind (;; ...
         :map perspective-map
         ("n" . nil)
         ("C-n" . persp-next)))
```
Or without `use-package`:
```elisp
(define-key perspective-map (kbd "n") nil)
(define-key perspective-map (kbd "C-n") 'persp-next)
```
Untested but should give you ideas about how to make it work. :) Feel free to follow up if you have trouble.

dams:
thanks !

ddoherty03:
I know this is closed, but why does using persp-mode-map rather than perspective-map not work.  When I use it in use-package, it binds keys to the global key map.  You say the map is persp-mode-map in the README.

ddoherty03:
Thanks.

PR #146, opened 2021-01-20
Add hint about how to remap keys

dams:
From #145


gcv:
Thanks for contributing to Perspective!

PR #151, opened 2021-03-03
add persp-switch-by-number command.

GZJ:
Add the method of quickly switching perspective through number keys, which is a bit similar to the way of i3wm switching workspace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ customize `persp-mode-prefix-key`. Additionally, creating a key binding for
 Here are the main commands defined in `persp-mode-map`:
 
 - `s` — `persp-switch`: Query a perspective to switch to, or create
+- `` ` `` — `persp-switch-by-number`: Switch to perspective by number, or switch quickly through number keys `1, 2, 3.. 0`
 - `k` — `persp-remove-buffer`: Query a buffer to remove from current perspective
 - `c` — `persp-kill` : Query a perspective to kill
 - `r` — `persp-rename`: Rename current perspective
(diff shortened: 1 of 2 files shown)


gcv (changes requested):
Good idea, I like it. Let's clean up the implementation, and I'll merge it in.

1. There's a lot of repetition in the `persp-switch-*` function family. Instead of using these, please pass the perspective number using a prefix argument (see documentation: https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html and https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Interactive.html). You'll need to change the `interactive` call in `persp-switch-by-number` to something like `(interactive "p\nnEnter perspective number:")` (untested; see interactive code documentation, https://www.gnu.org/software/emacs/manual/html_node/elisp/Interactive-Codes.html). This'll also work when the number of perspectives exceeds 10.
2. I'd prefer to get rid of the explicit `perspective-map` bindings to `(kbd "1")` etc., since the prefix-argument form will take care of this use case. If that's too many keystrokes for your taste, then I guess we can keep them, and you'll have to change each binding to something like `(lambda () (interactive) (persp-switch-by-number 1))`.
3. You can't rely on the `persp--modestring` frame parameter, because it's `nil` if `persp-show-modestring` is off. It may also change in the future. You also don't need a loop to achieve what you want. Just do something like `(nth (+ 1 num) (persp-names))` to access the perspective name you want by its number.
5. Please check your indentation. (Something like `crux-indent-defun` from the [Crux utility library](https://github.com/bbatsov/crux) might help.)

Side note: you don't need the loop at all, but for future reference, your loop code is too explicit. The Common Lisp `loop` macro is more appropriate here. I know it's arcane, so here's the gist of what you need (try this in ielm):
```elisp
(cl-loop for persp-name being the elements of (persp-names) using (index counter)
         collect (list (+ 1 counter) persp-name))
```

GZJ:
Ok, I changed the code.

gcv:
Merged. I took the liberty of improving `which-key` integration and added a small error-check. Thank you for contributing to Perspective.

PR #152, opened 2021-03-08
added persp-list-buffers function

titibandit:
I like to use C-x C-b to call the buffer-menu, that allows me to cleanup unused buffers.
But when I use perspective, I usually have 5 to 10 concurrent perspectives, which can amount to a lot of buffers in total, making C-x C-b almost unusable.
This is just a function that does the same, but filters for buffers in the current perspective.

There is not much to the function itself, so I can't really tell if it's just user customization, or if it has its place in perspective itself. Anyway, I find it very useful!
Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,11 @@ to use a Perspective-aware buffer switcher.
 in particular its `ido-switch-buffer` command, is automatically
 Perspective-aware when `persp-mode` is enabled.
 
+**buffer-menu**: Perspective provides a wrapper for
+[`buffer-menu`](https://www.gnu.org/software/emacs/manual/html_node/emacs/List-Buffers.html):
+`persp-buffer-menu`. When this function is called normally, it shows the buffer menu filtered by the current perspective. With a prefix argument, it shows the
+buffer menu of all the buffers in all perspectives.
+
 **`bs.el`**: Perspective provides a wrapper for
 [`bs-show`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Buffer-Menus.html):
 `persp-bs-show`. When this function is called normally, it shows a list of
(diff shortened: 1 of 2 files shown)


gcv:
Wow, I didn't think anyone used `buffer-menu` — I hated it so much when I started with Emacs (oh so many moons ago) that finding a different way to switch buffers by selecting from a list became my number one priority (number two: eliminating tilde backup file litter). In retrospect, the thing I truly hated was default display-buffer behavior. So, when I wrote Perspective-aware wrappers for buffer switching, I ignored that one. (See `persp-bs-show`, `persp-ibuffer`, `persp-ivy-switch-buffer`, `persp-counsel-switch-buffer`, and `persp-switch-to-buffer*`.)

That said, of course Perspective should have a native wrapper for it. A few requests for this PR.

1. Rename the function to `persp-buffer-menu` for consistency with the other functions in that family.
2. Move it to the same general area of the source file as the other functions which do this.
3. All the other buffer switching functions use the following pattern: without a prefix arg, they filter by perspective. With a prefix arg, they show the full buffer list. Please change your function to work the same way.
4. Update the Usage section of the README to include it.

titibandit:
Okay, I improved the function according to your recommendations.
Just a note, we called this buffer-menu, but emacs calls this also list-buffers. They are almost the same, but primarily mentioned as list-buffers in the emacs manual (https://www.gnu.org/software/emacs/manual/html_node/emacs/List-Buffers.html), but oh well.
Now I actually feel a bit dumb, since by doing this I've discovered the other options that exist, that seem better than this buffer-menu. But I guess it can't hurt to have that as well!

gcv:
I had no idea. :) For the sake of consistency, I decided to support both `list-buffers` and `buffer-menu` with their subtle differences. Also, since undoubtedly you won't be the last user to miss the buffer switchers in the README, I decided to recommend binding `C-x C-b` to `persp-list-buffers` in the `use-package` installation example.

PR #154, opened 2021-05-22
Add persp-ibuffer-set-filter-groups

TxGVNN:
Hi Maintainers,

I would like to add a feature to help users can group buffers by persp-name in ibuffer buffer. Inspired by https://github.com/purcell/ibuffer-projectile


gcv:
Looks like a good new feature. Merged to master. I had to add a few declarations to make the byte-compiler happy, please see https://github.com/nex3/perspective-el/blob/master/CONTRIBUTING.org#test-in-the-melpa-sandbox for future reference.

PR #155, opened 2021-06-07
Isolate xref--marker-ring between perspectives

TxGVNN:
Hi Maintainers,

I would like to make a PR to help users can isolate `xref--marker-ring` between perspectives. I've been using this feature for a long time.
Hope it can be useful for anyone.

Thanks!


gcv:
This is an excellent addition to Perspective. I decided to make it the default behavior, rather than opt-in. No need to activate it using the switch hook; it'll work out of the box. I also updated the documentation accordingly.

Thank you!

TxGVNN:
Oh, very well. I love it because I can reduce lines in my configurations =))
Thank you very much.

PR #158, opened 2021-07-08
persp-test-with-persp: use delete rather than delq to remove strings (equal vs eq)

mehw:
Hello,

Strings comparison requires `equal`, not `eq`, i.e. `delq` will not remove `"main"` (aka `persp-initial-frame-name`) from `'("main" "A")`.

Using `delq`, `persp-test-with-persp` will try to improperly kill a scratch buffer named `"*scratch* (main)"`, if `persp-names` returns a list like `'("main" "A")`.

### DIFF
--- a/test/test-perspective.el
+++ b/test/test-perspective.el
@@ -50,7 +50,7 @@ perspectives and open buffers."
      ;; get rid of perspective-specific *scratch* buffers first
      (mapc (lambda (persp)
              (kill-buffer (format "*scratch* (%s)" persp)))
-           (delq persp-initial-frame-name (persp-names)))
+           (delete persp-initial-frame-name (persp-names)))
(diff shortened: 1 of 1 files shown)


gcv:
Nice catch! How did you notice this?

mehw:
> Nice catch! How did you notice this?

Hi @gcv, I'm working on some PRs and, as [CONTRIBUTING.org](https://github.com/nex3/perspective-el/blob/46c7fdcdde074c4eb1b2c58ec2ee34df0efe509f/CONTRIBUTING.org) says, while running the tests I tumbled on it ;)

gcv:
Great! BTW, Perspective now uses GitHub Actions instead of Travis. I may have to manually enable it for you after you open your next PR.

PR #159, opened 2021-07-08
disabling persp-mode should not mix a buffer's header-line-format with its default (aka global) value

mehw:
Hello,

I noticed that disabling `persp-mode` while in a buffer with a locally modified `header-line-format` updates the default value of `header-line-format` to the buffer's local value... i.e. turning off `persp-mode` while in `*Buffer List*` (which has its own `header-line-format` set up) will set the header of buffers which do not have a locally `header-line-format` set up to the same header of `*Buffer List*`.

In this PR there's also an ert test to prove it... ;)


gcv:
Nice catch!

Issue #54, opened 2015-10-17
Add multiple buffers at once


alphapapa:
I use Helm, so when I tried `persp-add-buffer`, I selected several buffers with Helm, but only the first one was added.  It would be very helpful if you could add multiple buffers at once.  I was going to add all my org-mode buffers to an org-mode perspective, but doing it one-at-a-time...

zenspider:
I don't use helm, but patches are welcome.

Since nobody seems to be clamoring for this, it might be considered out of scope.

gcv:
Let's keep this one open for now. I actually ran into a situation yesterday when I wanted this feature and remembered this ticket.

gcv:
@alphapapa: I'm looking into this, and... I guess I'm not as proficient with Helm as I thought (which wasn't very).

How do I select a series of files in Helm and feed them to an arbitrary command, e.g., `persp-add-buffer`? I usually just end up in Helm's projectile integration browser, and I know how to mark several files, but hitting Tab gives me a canned list of commands. I don't see a way to invoke a command on the selection. I read on the wiki something about making eshell aliases, but though I can pick an alias somewhere in the command tree, it doesn't seem to do anything at all (or maybe I didn't define one correctly).

gcv:
With commit 0b693bdfdc135128c37bdcfaad0d57cc613564f4, `helm-buffers-list` now lists buffers in all perspectives when called with a prefix argument. It now has actions to add to the current perspective, and to remove buffers from the current perspective. It supports both actions with multiple marked buffers. If anyone has ideas about how to make Helm work directly from `persp-add-buffer`, let me know (either reopen this ticket or make a new one).

meliache:
I'd like to be able to add multiple buffers at once without helm, just using the generic `completing-read-multiple` (which I use via selectrum).

I was thinking about creating a new issue, but searched the closed issues first and found this. The issue title doesn't mention helm explicitly, so I think a more generic solution could be discussed.

Now that I think about it, it would be also useful to batch-remove buffers from a perspective. I haven't created my own `completing-read-multiple`-functions, but whenever packages use that out of the box (e.g. for tagging), I found it incredibly useful.

Instead of via completing-read-multiple, it is also convenient to add/remove multiple buffers from `ibuffer`. Here's what I came up with:
``` elisp
  (defun meliache/persp-ibuffer-remove-marked-buffers ()
    (interactive)
    (dolist (buffer (ibuffer-get-marked-buffers))
[...]

PR #162, opened 2021-09-09
Add commands to add/remove marked buffers to/from persp via ibuffer

meliache:
See [discussion](https://github.com/nex3/perspective-el/issues/54#issuecomment-915592609) in #54.
 
This allows to easily add and remove multiple buffers to/from a perspective. These functions have to be called from an ibuffer-buffer, after some buffers have been marked (e.g. via `m`).

I hope I have done the declare-function, the error-raising, autoloading etc. correctly. I'm used to writing elisp for my own emacs config, but haven't done much coding for packages yet, so these are things new to me, but I just tried to do the same as the existing functions.

## Potential TODO's

- [ ] Maybe add some default keybindings, a keymap or something like this for better ibuffer integration. Alternatively, just add an example configuration to the README.

- [ ] In any case, add some documentation for the functions to the README, e.g. in an _"Add/remove multiple buffers at once"_ section.

- [ ] Think whether we want to remove the marks after the buffers have been successfully added.

- [ ] When calling `persp-ibuffer-remove-marked-buffers` after `persp-ibuffer`, with an ibuffer buffer filtered to the current perspective, we could remove the removed buffers from the ibuffer view. This should not happen from an unfiltered ibuffer showing all perspective buffers.

- [ ] The commands work only in ibuffer, outside ibuffer they result in the error
  ```
  ibuffer-map-lines: Assertion failed: (derived-mode-p 'ibuffer-mode)
  ```
  Should we raise a custom user-error if the commands are called from outside ibuffer or is passing the assertion enough?

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1752,8 +1752,32 @@ restored."
         (pop-to-buffer ibuf)
         (ibuffer-update nil t)))))
 
+
+;;; --- modify perspectives from ibuffer
 
-;;; --- xref code
+;;;###autoload
+(defun persp-ibuffer-add-marked-buffers ()
+  "Add all marked buffers within ibuffer to the current perspective."
+  (interactive)
+  (unless (featurep 'ibuffer)
+    (user-error "IBuffer not loaded"))
+  (declare-function ibuffer-get-marked-buffers "ibuffer.el")
+  (dolist (buffer (ibuffer-get-marked-buffers))
+    (persp-add-buffer buffer)))
+
+;;;###autoload
+(defun persp-ibuffer-remove-marked-buffers ()
+  "Remove all marked buffers within ibuffer from the current perspective."
+  (interactive)
+  (unless (featurep 'ibuffer)
+    (user-error "IBuffer not loaded"))
+  (declare-function ibuffer-get-marked-buffers "ibuffer.el")
+  (dolist (buffer (ibuffer-get-marked-buffers))
+    (persp-remove-buffer buffer)))
+    ;; TODO: if the current ibuffer is filtered to the current perspective, update
+    ;; the view to not show the removed buffers
+
+ ;;; --- xref code
 
 ;; xref is not available in Emacs 24, so be careful:
 (when (require 'xref nil t)
--- a/perspective.el
+++ b/perspective.el
@@ -1769,6 +1769,12 @@ restored."
   (persp--ibuffer-function-on-marked #'persp-add-buffer))
 
 ;;;###autoload
+(defun persp-ibuffer-set-marked-buffers ()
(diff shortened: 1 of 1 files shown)


gcv:
I tried your change, it looks good. I'm okay with shipping this without keybindings and just a README note.

With regard to updating the ibuffer buffer after removing buffers from the perspective, I tried this modification to your code, and it seems to work:

```elisp
(defun persp-ibuffer-remove-marked-buffers ()
  "Remove all marked buffers within ibuffer from the current perspective."
  (interactive)
  (unless (featurep 'ibuffer)
    (user-error "IBuffer not loaded"))
  (defvar ibuffer-maybe-show-predicates)
  (declare-function ibuffer-get-marked-buffers "ibuffer.el")
  (declare-function ibuffer-update "ibuffer.el")
  (dolist (buffer (ibuffer-get-marked-buffers))
    (persp-remove-buffer buffer))
  (let ((ibuffer-maybe-show-predicates (append ibuffer-maybe-show-predicates
                                               (list #'persp-buffer-filter)
                                               ido-ignore-buffers)))
    (ibuffer-update nil t)))
```

It's a bit ugly because it copy-pastes the `ibuffer-maybe-show-predicates` binding from `persp-ibuffer`. Maybe a `persp--ibuffer-maybe-show-predicates-helper` function could be added that returns the necessary list, and then `persp-ibuffer` and `persp-ibuffer-remove-marked-buffers` could both use it for the binding. I'm open to hearing about a nicer way to refactor it.

gcv:
I just realized my code will enforce `persp-ibuffer` behavior even when used in non-perspective-specific ibuffer. :(

There are ways to hack around this. For example: have `persp-ibuffer` set `persp--ibuffer-called-with-persp` flag in a buffer-specific variable on the `*IBuffer*` buffer, and then check that flag from `persp-ibuffer-remove-marked-buffers`.

gcv:
Let's do one other thing to reorganize the code a bit. Let's take the entire "ibuffer filter group code" section along with your new code, rename it to "ibuffer integration", and move it all next to the "Helm integration section" (either before or after, doesn't matter). That way all "integration" code sits together. Then move the `persp-ibuffer` function into the new section. Feel free to do that as part of this PR.

Let me know when you finish any other refactoring you plan to tackle for the purposes of this PR, and I'll take another look and merge.

meliache:
Sorry for ghosting this PR for so long, back then I got busy with other things and then also stopped using perspective.el because I found it to not be necessary for my workflow. If you still feel like the functionality that I suggested in this PR as a way to address #54 via ibuffer is useful, please implement it in a separate PR, though feel free to reuse the code. Though looking at the diff the so far added code in this PR was very minor anyway.

PR #170, opened 2021-11-21
Do not require which-key and xref

aaronjensen:
Loading persp is the slowest part of my emacs startup. This helps a bit by not requiring things that I can require later, as needed. I do not know if there is a reason not to do this, but it seems better to respond to things loading rather than force their loading.

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -312,9 +312,7 @@ Run with the activated perspective active.")
 (define-key perspective-map (kbd "9") (lambda () (interactive) (persp-switch-by-number 9)))
 (define-key perspective-map (kbd "0") (lambda () (interactive) (persp-switch-by-number 10)))
 
-(declare-function which-key-mode "which-key.el")
-(when (fboundp 'which-key-mode)
-  (require 'which-key)
+(with-eval-after-load 'which-key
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -312,22 +312,23 @@ Run with the activated perspective active.")
...
-      "5" "switch to 5"
-      "6" "switch to 6"
-      "7" "switch to 7"
-      "8" "switch to 8"
-      "9" "switch to 9"
-      "0" "switch to 10")))
+(with-eval-after-load 'which-key
+  (declare-function which-key-mode "which-key.el")
+  (when (fboundp 'which-key-mode)
+    (require 'which-key)
```
gcv, line 318: Please verify that this check  and `require` call is still necessary with the use of `with-eval-after-load`.

gcv (changes requested):
I like the idea of streamlining startup! If the check for which-key isn't needed, it should make your diff cleaner.

perspective.el
aaronjensen: Ah, good call. Thanks!

gcv:
Thanks, merged.

PR #172, opened 2021-12-12
Fix restore of multibyte buffer names when loading state from file.

intramurz:
Hi!

When the state is saved to a file, the multibyte-encoded names of the buffers are saved as they are, but when the state is loaded, they are restored as backslash sequences. Thus, the configuration of the buffers changes on save/restore.  Such names look terrible in the mode-line, some commands do not work correctly with them.
I suppose if someone gave a multibyte name to the buffer before the state was saved, it works after the restore as well.
Or if you have some reason to act this way, maybe I could add a custom option to control this? 

Thank you

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -1863,7 +1863,7 @@ restored."
         (frame-count 0)
         (state-complete (read
                          (with-temp-buffer
-                           (insert-file-contents-literally file)
+                           (insert-file-contents file)
                            (buffer-string)))))
     ;; open all files in a temporary perspective to avoid polluting "main"
     (persp-switch tmp-persp-name)


perspective.el
```diff
@@ -1863,7 +1863,7 @@ restored."
         (frame-count 0)
         (state-complete (read
                          (with-temp-buffer
-                           (insert-file-contents-literally file)
```
gcv, line 1866: @zenspider: It looks from `git blame` that you changed `insert-file-contents` to `insert-file-contents-literally` here in December 2019. Do you remember why? I can confirm that it causes the problem @intramurz's PR is trying to fix.

zenspider: Can't say I remember... but looking at the differences in the two, I must have either been having problems with the coding conversion or (my guess) file-name handlers, which `*-literally` overrides with `inhibit-file-name-handlers` ... which just seems backwards to me...

Go ahead and flip it. I'll squock it bites me.

gcv:
@intramurz: Merged. Thank you for the bug report and the PR. Let's see if this breaks something else. 😆

PR #176, opened 2022-03-06
Add consult integration

ivan-pop:
This adds new variable `persp-consult-source` which is source that can be used
with `consult-buffer-sources` and instruction to README on how to use it

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,17 @@ show buffers in all perspectives. In addition, Perspective adds actions to
 to the prefix-argument version) and to remove buffers from the current
 perspective.
 
+**Consult**: Perspective provides `persp-consult-source` source that will list
+buffers in current perspective. You can hide default buffer source
+and add `persp-consult-source` to `consult-buffer-sources` for consult
+to only list buffers in current perspective, note that you can access
+list of all buffers by filtering with `b` key. Example configuration:
+
+```emacs-lisp
+(consult-customize consult--source-buffer :hidden t :default nil)
+(add-to-list consult-buffer-sources persp-consult-source)
+```
+
 **Ivy / Counsel**: Perspective provides two commands for listing buffers using
 Ivy and Counsel: `persp-ivy-switch-buffer` and `persp-counsel-switch-buffer`.
 When these functions are called normally, they show a list of buffers filtered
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -246,6 +246,17 @@ show buffers in all perspectives. In addition, Perspective adds actions to
...
 
+**Consult**: Perspective provides `persp-consult-source` source that will list
+buffers in current perspective. You can hide default buffer source
+and add `persp-consult-source` to `consult-buffer-sources` for consult
+to only list buffers in current perspective, note that you can access
+list of all buffers by filtering with `b` key. Example configuration:
+
+```emacs-lisp
+(consult-customize consult--source-buffer :hidden t :default nil)
+(add-to-list consult-buffer-sources persp-consult-source)
```
gcv, line 257: `add-to-list` takes a symbol, so that should be `(add-to-list 'consult-buffer-sources persp-consult-source)`.

README.md
```diff
@@ -246,6 +246,17 @@ show buffers in all perspectives. In addition, Perspective adds actions to
 to the prefix-argument version) and to remove buffers from the current
 perspective.
 
+**Consult**: Perspective provides `persp-consult-source` source that will list
+buffers in current perspective. You can hide default buffer source
+and add `persp-consult-source` to `consult-buffer-sources` for consult
+to only list buffers in current perspective, note that you can access
+list of all buffers by filtering with `b` key. Example configuration:
```
gcv, line 253: How do I filter with <kbd>b</kbd>? For me, that just filters to buffers containing the letter `b`. (I do use Consult, but I'm not an expert.)

gcv (changes requested):
Thanks! Just a couple of nitpicks before merging.

ivan-pop:
@gcv Just pushed fixes. Regarding `consult--buffer-state` it is defined in consult using `consult--define-state` macro

ivan-pop:
Should i maybe wrap config part in README in `with-eval-after-load` too?

gcv:
> Should i wrap config part in README in `with-eval-after-load` too?

I don’t know. Part of me thinks it’ll be nicer and more user-friendly to do it, but that kind of automatic setup seems counter to the “let the user decide” spirit of Vertico and Consult. Feel free to do whichever you think is right and I’ll merge it.

ivan-pop:
I agree that it is better to let user decide it, lets go without it

gcv:

PR #177, opened 2022-04-04
Wrap with-perspectives' BODY arg in a progn and return it.

NicholasBHubbard:
This PR wraps `with-perspective`s BODY arg in a `progn` and then returns the value.

I don't see any reason for this not to be the behavior.


NicholasBHubbard:
I should add that this aligns with the behavior of `with-current-perspective`.

gcv:
Looks good to me. Let’s see if anything breaks. :)

PR #184, opened 2022-05-21
feat: compatible with xref in emacs29

TxGVNN:
`xref--marker-ring` will be replaced by `xref--history` in emacs-29 (Or just updating xref version to latest).
So I make PR to make the perspective still support xref.

Thanks!


gcv:
Thanks!

PR #187, opened 2022-07-05
Remove needless quote from the choice value

syohex:
When you set its value via `customize`, Emacs generates following code into your init.el

```
(custom-set-variables
 '(persp-show-modestring ''header))
```


gcv:
Thank you, nice catch.

PR #188, opened 2022-07-19
Feature: persp-purge-initial-persp-on-save - purges the inital persp upon save

titibandit:
This is a small feature that I use since quite a few months:
I basically use my initial perspective as a "scratchpad" for quickly editing files. I do so, so I don't clutter my "project" perspectives with buffers that are very temporary and that I might forget to kill right away, plus, they usually don't belong to any of my longer-term perspectives anyway.

With this feature, I can basically open all the temporary buffers in my initial persp and then upon using `persp-state-save`, the inital persp will get "purged", meaning all buffers, besides the ones that have names that match any regexp of the list `persp-purge-initial-persp-on-save-exceptions` will be killed. For the moment, I just have my `init.el` there, because it's the only file I consistently need to edit daily :)

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,10 @@ customize`). The following are likely to be of most interest:
   the modeline).
 - `persp-modestring-short`: When set to `t`, show a shortened modeline string
   with only the current perspective instead of the full list. Defaults to `nil`.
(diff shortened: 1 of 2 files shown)


gcv:
Thanks! Merged.

PR #201, opened 2023-09-26
Allow manual sorting of perspectives

sipraga:
Users can select a manual sorting method and then rearrange the list of perspectives using newly added functions pers-sorting-move-left/right.


gcv:
Thanks, this looks like a useful feature. Unfortunately, the implementation does not work for me.

First, the change to the `perspective` struct adds a field called `sorted-time`, but all the code references `persp-sorting-time`.

Second, even if I change `persp-sorting-time` to `persp-sorted-time`, calls to `(persp-names)` still fail.

PR #215, opened 2025-02-11
persp-reactive-buffers: preserve recent visited order of buffers

blahgeek:
Previously, after switching to a perspective, the recent-visit order of buffers will be changed because `persp-reactive-buffers` revisits buffers based on `persp-current-buffers` which is in a fixed order. With this PR, `persp-reactive-buffers` revisits the buffers based on `(buffer-list)`

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -663,11 +663,11 @@ buffer called \"*scratch* (NAME)\"."
 Returns BUFFERS with all non-living buffers removed.
 
 See also `other-buffer'."
-  (cl-loop for buf in (reverse buffers)
-           when (buffer-live-p buf)
-           collect buf into living-buffers
+  (cl-loop for buf in (reverse (buffer-list))
+           when (member buf buffers)
+           collect buf into result-buffers
            and do (switch-to-buffer buf)
-           finally return (nreverse living-buffers)))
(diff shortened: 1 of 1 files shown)


perspective.el
```diff
@@ -663,11 +663,11 @@ buffer called \"*scratch* (NAME)\"."
 Returns BUFFERS with all non-living buffers removed.
 
 See also `other-buffer'."
-  (cl-loop for buf in (reverse buffers)
-           when (buffer-live-p buf)
-           collect buf into living-buffers
+  (cl-loop for buf in (reverse (buffer-list))
+           when (member buf buffers)
```
gcv, line 667: Let's keep the check for `buffer-live-p` here (`(and (buffer-live-p buf) (member buf buffers))` or something like it). It might be overkill, but Perspective had a lot of "dead buffer" bugs over the years and I'd rather be more cautious.

gcv (changes requested):
Looks good, but please consider the `buffer-live-p` check I noted in the comment.

perspective.el
blahgeek: Thanks. Updated.

gcv:

PR #218, opened 2025-02-18
Fix name in `.mailmap`

LV:
Found the commit title "Fix my name" interesting and the fact that it was 10 years old, so I looked at it and noticed that `.mailmap` had incorrect names and it bothered me too much not to create a PR.

Sorry if this is weird.


gcv:
I never paid attention to this file before, but now that I’ve looked at it, I think it’s supposed to be the way it is: it’s supposed to create an alias from the old name to the new name for Git tracking purposes. I looked it up here: https://lukasmestan.com/using-mailmap-in-git-repository/

If I understand it correctly, applying your patch will cause the name change to be ignored in old Git history.

LV:
My bad, disregard this.

PR #220, opened 2025-10-31
Trust the default perspective scratch buffer contents

elken:
Without this, by default trying to interact with a new perspective
puts the "Disabline elisp-flymake-byte-compile due to untrusted
content" message to popup in the minibuffer creating an annoying UX.

Regular scratch should trusted in Emacs[1][2] but that doesn't propogate
here

[1] https://lists.gnu.org/archive/html/emacs-devel/2025-02/msg01023.html
[2] https://github.com/emacs-mirror/emacs/commit/8b6c6cffd1f7

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -312,7 +312,12 @@ Create it if the current perspective doesn't have one yet."
         (when (and (zerop (buffer-size))
                    initial-scratch-message)
           (insert (substitute-command-keys initial-scratch-message))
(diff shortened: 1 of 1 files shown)


gcv:
I just read the mailing list thread, and have more questions than answers from it. Is there maybe an official function that we can call that makes a new scratch buffer we can call that does the official right thing? I don't feel comfortable setting such a sensitive setting on the user's behalf.

PR #228, opened 2026-06-23
Improve prompting of persp-remove-buffer by naming current/default buffer

zenspider:
Lets you know what you're defaulting to if you hit return as-is.
Otherwise, you could have a tangle of buffers open at the time and not
realize the context you were just coming from.


gcv (approved):
Looks fine to me, feel free to merge. I never noticed because my vertico configuration shows the current buffer as the first thing in the completion.

PR #230, opened 2026-06-24
README: fix Buffer switching link

haplo:
I found this broken link.

Thanks for perspective!


gcv:
Good catch, thank you!

PR #232, opened 2026-06-27
feat(filter): ido-ignore-buffers functions support

0jset:
add ability to use functions for filtering of buffers like in ido

add dependency on dash

### DIFF
--- a/perspective.el
+++ b/perspective.el
@@ -244,15 +244,16 @@ selected window's buffer current instead of the caller's buffer."
          (set-frame-parameter nil 'persp--last last-persp-cache)
          result))))
 
-(defun persp--make-ignore-buffer-rx ()
+(defun persp--ignore-buffer-p (name)
+  "Check if buffer with NAME should be ignored based on `ido-ignore-buffers'."
   (defvar ido-ignore-buffers)
-  (if ido-ignore-buffers
-      ;; convert a list of regexps to one
-      (rx-to-string (append (list 'or)
-                            (mapcar (lambda (rx) `(regexp ,rx))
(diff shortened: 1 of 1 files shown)


gcv:
I’m not opposed to adding functions for buffer filtering, but I will not approve a dependency on dash. Perspective tries very hard to be free of dependencies, and I want to continue supporting fairly old versions of Emacs.

0jset:
> I want to continue supporting fairly old versions of Emacs.

[Dash requires emacs 24,](https://github.com/magnars/dash.el/blob/master/dash.el#L8) and [perspective-el requires emacs 24.4](https://github.com/nex3/perspective-el/blob/master/perspective.el#L9)

gcv:
Fair point! But I still want to keep Perspective dependency-free. I think your code is pretty easy to rewrite using just the operations available in basic Elisp.

0jset:
@gcv I have rewritten the logic and added a few more updates. Please review the latest changes when you get a chance.

gcv:
Thanks for removing dash! Two more things before I merge:

  - Please drop the unrelated indentation changes so the functional diff stays focused.
  - The BS predicates receive buffer objects, but `persp--ignore-buffer-p` expects a buffer name. Please pass `(buffer-name buf)` there and ensure the all-perspectives predicate is an actual callable function.

0jset:
@gcv Fixed it:
  - Dropped the commit with indentations.
  - BS predicates now use `(persp--ignore-buffer-p (buffer-name buf))` 
  - squashed commits

gcv:

